PR #28, opened 2020-06-24
Add Invitations

sboisvert:
Goal:
We need to be able to send invitations by email to front line healthcare workers who will then be allowed to signup to generate codes.

This PR includes:

- The Django-invitations app
- The Django-invitation templates inside profiles/templates/invitations/* (two have been overwritten the others will be overwritten later)
- Additional Email_* settings that are required to be set in the environment or .env file
- Enforces a 12 character minimum password length
- New Signup form that extends HealthcareBaseForm (includes verification that the email was sent an invite)
- New html and txt email template (of note, gmail doesn't support inline SVG so we'll need to change that.
- Adds test to ensure email is prepopulated

image
image
image


sboisvert:
@pcraig3 I'm still manually testing this but I think it's ready for review. I'm going to add some unit tests to it but not sure if we want to merge it now without the tests to make it easier for future merges

sboisvert:
I'm going to work on tests for this in a separate PR

profiles/forms.py
```diff
@@ -55,13 +55,59 @@ def __init__(self, *args, **kwargs):
...
+    def clean_password2(self):
+        password1 = self.cleaned_data.get('password1')
+        password2 = self.cleaned_data.get('password2')
+
+        if password1 and password2 and password1 != password2:
+            raise ValidationError(_("Password don't match"))
+
+        #We probably don't actually need this here and can rely on AUTH_PASSWORD_VALIDATORS
+        if len(password2) < 12:
+            raise ValidationError(_("Password must be at least 12 characters"))
```
pcraig3, line 99: I agree, let's remove the "No numeric password" rule and up the password field in settings to 12.

- https://github.com/cds-snc/covid-healthcare-portal/blob/main/portal/settings.py#L121

profiles/views.py
```diff
@@ -2,19 +2,28 @@
...
+    if request.method == 'POST':
+        f = SignupForm(request.POST, initial = {"email" : request.session.get('account_verified_email', None)})
+        if f.is_valid():
+            f.save()
+            messages.success(request, 'Account created successfully')
+            return redirect('start')
+ 
+    else:
+        prepopulate = {}
+        prepopulate["email"] = request.session.get('account_verified_email', None) if request.session.get('account_verified_email', None) else None
```
pcraig3, line 22: Do you know if the 'account_verified_email' key is always going to be in the session, or will it only be there if the value we want exists?

The second argument to the `get` method is what gets returned if the key doesn't exist, but it won't be returned if the key _does_ exist.

```py
dic = {'a': False}
print(dic.get('a', 'No'))) # prints False
print(dic.get('b', 'No'))) # prints 'No'
```

So if you know the key won't exist unless the value is there, you can just leave this as:
`request.session.get('account_verified_email', None)`

But if the key might be there with an empty value or something, you can do:
`request.session.get('account_verified_email') or None`

The way it's written now is really confusing.

profiles/views.py
```diff
@@ -2,19 +2,28 @@
...
+            return redirect('start')
+ 
+    else:
+        prepopulate = {}
+        prepopulate["email"] = request.session.get('account_verified_email', None) if request.session.get('account_verified_email', None) else None  
+        f = SignupForm(initial=prepopulate)
+ 
+    return render(request, 'profiles/signup.html', {'form': f})
 
-class SignUp(generic.CreateView):
```
pcraig3, line 14: I was going through `learndjango.com` last week and there's an article on there about [preferring 'Class-based views' to 'Function-based views'](https://learndjango.com/tutorials/django-best-practices-function-based-views-vs-clas).

Basically, it makes the argument that if your view isn't doing anything really custom, to prefer a Class-based view. It seems like it's newer Django and results in less repeat code over time.

I would prefer sticking with Class-based views as far as we can.

For form handling with CBVs, there's pretty good documentation on it: https://docs.djangoproject.com/en/3.0/topics/class-based-views/generic-editing/#form-handling-with-class-based-views

To pre-populate the form, the old `UserEdit` view did some initial form population with `get_initial`: https://github.com/cds-snc/covid-healthcare-portal/blob/012b960b9519167387961a6c5a90ff2a060b1331/portal/profiles/views.py#L52

I know it works now, so we can merge as-is and update later. Happy to do refactoring too, and you can review.

pcraig3:
One last thing I'll add is that you can update the [`signup.html`](https://github.com/cds-snc/covid-healthcare-portal/blob/main/profiles/templates/profiles/signup.html) template to bring in the new form styling and error messaging. 

I already did this for the PasswordReset form so it will be pretty straightforward: https://github.com/cds-snc/covid-healthcare-portal/commit/7ac342d1a3e10e8a06d7ffbba9d6bf92ba2d9322#diff-d070b5025129bb20e939468414da2952

portal/settings.py
```diff
@@ -175,7 +175,18 @@
...
+
+# Email setup
+EMAIL_BACKEND = os.getenv("EMAIL_BACKEND")
+DJANGO_SECRET_KEY = os.getenv("DJANGO_SECRET_KEY")
+EMAIL_HOST = os.getenv("EMAIL_HOST")
+EMAIL_PORT = int(os.getenv("EMAIL_PORT"))
+EMAIL_HOST_USER = os.getenv("EMAIL_HOST_USER")
+EMAIL_HOST_PASSWORD = os.getenv("EMAIL_HOST_PASSWORD")
+email_use_tls = os.getenv("EMAIL_USE_TLS")
+EMAIL_USE_TLS = True if email_use_tls == "True" else False
```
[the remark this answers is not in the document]
LouisStAmour: https://docs.djangoproject.com/en/3.0/ref/settings/#email-use-tls but it's concerning that apparently Django still supports setting both EMAIL_USE_SSL and EMAIL_USE_TLS as if TLS hasn't replaced every use of SSL I can think of...

profiles/views.py
LouisStAmour: Just an aside, but I suddenly want "Classy-Class-Based-Views" in my next IDE... https://ccbv.co.uk/projects/Django/3.0/django.contrib.auth.views/PasswordChangeDoneView/

sboisvert:
@pcraig3 I think you need to re-approve as a code-owner

PR #33, opened 2020-06-25
Remove user "language" attribute + implement CDS/Canada.ca login form 

pcraig3:
This PR does two main things

1. Removes "language" attribute from our users
2. Implements the login form with CDS/Canada.ca styles

### 1. Remove `language` attribute from HealthcareUser

As @brianhendrick1 rightly pointed out when we were chatting, users don't need to "save" a language since it's already on the site at the top. They can just change if they want and it'll remain set in their session.

Removed "language" from our HealthcareUsers and added a migration

### 2. Implements the login form with CDS/Canada.ca styles

The Django default login form uses [AuthenticationForm](https://github.com/django/django/blob/f3e5fba7c5d6fbfa8d8d0694eb4ab311f874de5d/django/contrib/auth/forms.py#L173), which I was able to import and subclass to make a few tweaks to. 

Added CDS form styling, spacing rules, error message styling, and an `errorlist` component at above the `h1` if there are errors. 

Inside of the login template, I'm using Django's `form` attribute to [render errors and form elements](https://docs.djangoproject.com/en/3.0/topics/forms/#rendering-fields-manually) in a general way, so we can reuse the same code in an `include` for future forms. Once we get the signup form working how we expect, we can use this same styling for that as well.

## Screenshots

|                                  | before | after |
|----------------------------------|--------|-------|
| blank login form                 |   Screen Shot 2020-06-25 at 9 56 35 AM   |  Screen Shot 2020-06-25 at 9 56 24 AM   |
| login form with validation error | Screen Shot 2020-06-25 at 9 57 06 AM  | Screen Shot 2020-06-25 at 9 57 03 AM  |


profiles/templates/registration/login.html
```diff
@@ -1,13 +1,52 @@
...
+    <h2 class="errorlist--heading">
+      {% trans "Please correct the errors on the page" %}
+    </h2>
+    <ol class="errorlist">
+      {% for key,value in form.errors.items %}
+        <li>
+        {% if key == '__all__' %}
+          <a href="#form--{{ request.resolver_match.url_name }}">
+        {% else %}
+          <a href="#id_{{ key|escape }}">
```
sboisvert, line 21: I thought the templates auto-escaped unless you did `| safe`

profiles/templates/registration/login.html
```diff
@@ -1,13 +1,52 @@
...
+        {% endif %}
+          {{ value|first|escape }}
```
sboisvert, line 23: same as above

profiles/templates/registration/login.html
pcraig3: Good catch!

Tested putting unsafe characters in error messages and they
display the same with or without "|escape", so we don't need it.

The only way to interpret HTML tags and such is to use "|safe".

More details here: https://github.com/cds-snc/covid-healthcare-portal/pull/33/commits/d5ceeb039c6accd76eb8d5f2f08dbd35d4af336e

PR #38, opened 2020-06-25
Shows the correct length of the code generated

PierrePaul:
Showing the real length of the code generated.
There was a few changes needed to make the translation works in the Docker container, but I haven't included them in this merge request for now. Let me know if you would like them at the same time.

Screenshot_2020-06-25 Fournir le code au patient

Screenshot_2020-06-25 Provide patient with code


pcraig3:
Hello! 

Our design for this page is changing still, but we're still keeping the "num" digit code aspect of it, so this seems like a good addition. 

The few changes you're mentioning in the Docker container, are you talking about adding a "compile" step to `./entrypoint` or is there something else?

PierrePaul:
The docker container doesn't have the `gettext` utilities required to `makemessages` or `compilemessages`.

PierrePaul:
> We accept an 8-digit code but then add an empty space in the middle so that it's more readable. Looking at the string length give you `9` because of the space in the middle, so it's probably better to keep the number hardcoded for now.

You're right! I missed that. I can fix this easily if you want. If you're more comfortable with hard coded, I totally understand too.

pcraig3:
Sorry about the approval and then the retraction. I didn't figure out about the miscount problem until I had sent the approval along. If you want to change this so it's not dynamic (or ignores the space), I'll re-approve the PR.

> The docker container doesn't have the `gettext` utilities required to `makemessages` or `compilemessages`.

Ah, interesting. If you want to submit a PR for that, that would be super.

sboisvert:
@PierrePaul Merci pour le PR! My initial reaction is to suspect that since the 8 digit code won't change in size that we can keep it hardcoded. Just really to keep things as simple as possible, what do you think?

PierrePaul:
> @PierrePaul Merci pour le PR! My initial reaction is to suspect that since the 8 digit code won't change in size that we can keep it hardcoded. Just really to keep things as simple as possible, what do you think?

It's your call really, right now the code looks like `code|cut:" "|length` so I wouldn't call this really complicated. That exact wording might not even be necessary in the future.

pcraig3:
Gonna merge, as I am working on this file now.

PR #42, opened 2020-06-26
Readme and contribution guide file updates

sboots:
Minor formatting change to the Readme, plus contribution files for consistency with the other exposure notification repositories.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,42 @@
+[La version française suit.](#contribution)
+
+# Contributing
+
+Thank you for considering contributing to COVID Shield!
+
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,42 @@
+[La version française suit.](#contribution)
+
+# Contributing
+
+Thank you for considering contributing to COVID Shield!
```
sboisvert, line 5: Do we want to change this since this isn't tied to the original COVID Shield apps? Or should we just wait for an official name?

sboots: Great point @sboisvert! My vote is, leave these as-is for now and then bulk-change them as needed once there's an official name.

sboots:
Thanks @sboisvert! 🙌

PR #44, opened 2020-06-26
Add MIT LICENSE

pcraig3:
Open to all, but you can't sue us.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 Canadian Digital Service – Service numérique canadien
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


pcraig3:
Good question.

For all of our current repos we just used the English version of the MIT license. 

This stackexchange question implies that translating the license isn't really useful: https://opensource.stackexchange.com/questions/6077/license-details-in-languages-other-than-english-american/6079#6079

> Translations of the licenses may be helpful for understanding, but are not a legally relevant document. If an official translation exists you could additionally link to that. You should not write or commission a translation of your own. That exposes you to unnecessarily liability if you mistranslate something, and it may actually violate the license of the license. 

So we're not worried about a French translation of the MIT license. 

Other than that, we translate our READMEs and other project documentation as a standard practice

PR #45, opened 2020-06-26
Update the landing, start, and code pages to be in more line with Figma

pcraig3:
Updates the app name from "StopCOVID" to "COVID Alert", plus the content on most of the other pages.

### /landing 

| before | after |
|--------|-------|
|  Screen Shot 2020-06-26 at 2 24 44 PM   |   Screen Shot 2020-06-26 at 2 24 57 PM  |

### /start

| before | after |
|--------|-------|
|  Screen Shot 2020-06-26 at 2 25 23 PM  | Screen Shot 2020-06-26 at 2 25 21 PM    |

### /code 

| before | after |
|--------|-------|
| Screen Shot 2020-06-26 at 2 25 39 PM    |  Screen Shot 2020-06-26 at 2 25 34 PM   |


profiles/static/scss/_base.scss
```diff
@@ -143,3 +143,7 @@ li:not(:last-of-type) {
   font-size: 1.5em;
   border-radius: 10px;
 }
+
+.-mt10 {
```
LouisStAmour, line 147: (Typing from mobile so I can’t see the entire file) 

I can’t say I’m familiar with every CSS naming preference, but is there a reason this starts with a hyphen and is somewhat obscurely named? If it matches existing patterns, sure, but if this is new, I’d use a more descriptive or semantic name?

pcraig3: it's similar to other libs like tailwind where you can add utility-based CSS classes: https://tailwindcss.com/docs/margin/#negative-margins

But it's a fair point, I can change this to be more descriptive.

PR #47, opened 2020-06-26
Add expiry date to code page in en/fr

pcraig3:
Added an "expires" string to the '/code' page with the current time, in both English and French formatting.

The reason I'm handling the time formatting in the template is that modifying the timezone on the server caused us to miss out on time localization. In addition, the template `date` filter options are are more numerous and flexible.

Using EST time for this PR, changing timezones is outside the scope of the story.
The short version is that to get a user's timezone we either have to use JS to guess the locale, or get the user to input their timezone in some kind of settings menu.

Generally the less we have to deal with this stuff, the better.

Sources:
- https://docs.djangoproject.com/en/3.0/topics/i18n/timezones/
- https://tommikaikkonen.github.io/timezones/
- https://docs.djangoproject.com/en/3.0/ref/templates/builtins/#date
- https://stackoverflow.com/questions/10801397/system-date-formatting-not-using-django-locale

## Screenshots

| before | after |
|--------|-------|
|  Screen Shot 2020-06-26 at 6 42 47 PM   |  Screen Shot 2020-06-26 at 6 42 44 PM    |


LouisStAmour:
Thought ... We could do (in JS):

```js
`Expires: ${(new Date(Date.now()+10*60*1000)).toLocaleString('fr-CA', { weekday: 'long', year: 'numeric', month: 'long', day: 'numeric', hour: 'numeric', minute: 'numeric', second: 'numeric', timeZoneName: 'short' })}`
```

And even pass through the server's UTC timestamp by having `Date.now()+10*60*1000` (or however close we can get) computed server-side, because UTC time should be an accurate representation of what time it will be when this expires (with odd-but-accurate effects during leap hours):

```js
`Expires: ${(new Date(1593236567072)).toLocaleString('fr-CA', { weekday: 'long', year: 'numeric', month: 'long', day: 'numeric', hour: 'numeric', minute: 'numeric', second: 'numeric', timeZoneName: 'short' })}`
```

In the above, fr-CA can be swapped for en-CA, though "Expires: " would have to be translated like other strings.

It's worth keeping in mind that when something is predicted to happen in x minutes, or is a log of what happened x minutes ago, you can use UTC timestamps (or UNIX timestamps) to represent these instants in time. Where you can't use UNIX timestamps is when humans are trying to arrange a time to meet in a certain location's time zone (because time zones vary over time and place), but when minutes get leap seconds added to them, the duration of a minute changes in every time zone -- such that generally speaking, 10 minutes from now in UTC is equivalent to 10 minutes from now in a local time zone. :)

Through this, you could create a JS-based countdown clock, change the page on its expiry, etc.

PierrePaul:
Just throwing my 2 cents, storing the date in UTC and using JS to show the date in the correct timezone (either user preference or browser default) are usually the most reliable way to go.

Moment.js is easy to work with, but big. There are other alternatives like date-fns with good timezone and i18n support ( https://github.com/you-dont-need/You-Dont-Need-Momentjs ).

There is also the possibility of storing UTC, then using `<time>` tag, but if I recall correctly, it's not supported in internet explorer 11. Also, using `<time>`, a user could not change manually the timezone in his profile, the user is stuck with the timezone exposed by his browser.

sboisvert:
I think we want to go quick and simple on this, I don't think we need to worry about exact precision with leap seconds and etc.
I think we can assume that the computer they are using has the correct timezone, or at least one they will know and recognize, so if we say EDT and their computer is in EDT but they are currently in CDT, they are probably aware of this.
@PierrePaul do you want to take a go at using date-fns for this?

PierrePaul:
> @PierrePaul do you want to take a go at using date-fns for this?

Yes, on it.

pcraig3:
Gonna merge as-is for now and then later we can revisit.

pcraig3:
oops, can't merge until I get another approval

PR #49, opened 2020-06-28
Update README.md

sboisvert:
Add a bit of info to the "before you commit" section

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,14 @@ Run `python manage.py compilemessages` to compile the translations so that Djang
 
 For more complete documentation refer to the [Django Translation](https://docs.djangoproject.com/en/3.0/topics/i18n/translation/#translation) docs.
 
+## Before you commit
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -110,6 +110,14 @@ Run `python manage.py compilemessages` to compile the translations so that Djang
 
 For more complete documentation refer to the [Django Translation](https://docs.djangoproject.com/en/3.0/topics/i18n/translation/#translation) docs.
 
+## Before you commit
+1. Ensure that you have a commit message that explains all the changes.
+2. Run:
+    a) `python manage.py makemessages -l fr`
+    b) `python manage.py compilemessages`
```
pcraig3, line 117: We don't actually need to do this unless we have French translations to enter. Doing this helps us if we have French translations to add, or if we want to clear out a bunch of english translation keys that aren't being used anymore.

PR #51, opened 2020-06-28
Set "production" environment variables

pcraig3:
Because `DJANGO_DEBUG` and `DJANGO_ENV` weren't being looked for in the Dockerfile, they weren't being passed into the container, so the `is_prod` var was never getting set.

This PR:
- updates the Dockerfile so that we can pass in env vars
- Removes the env vars from app.json which aren't actually getting passed to the container (I checked)
- sets the `SECURE_PROXY_SSL_HEADER` setting, without which we get infinite redirects on Heroku
  - Note that if we switch from Heroku, we might need to delete this setting


Dockerfile
```diff
@@ -5,13 +5,19 @@ FROM python:3.8
...
 RUN apt-get update && apt-get install -y --no-install-recommends \
-		gettext \
+	gettext \
 	&& rm -rf /var/lib/apt/lists/*
 
 # Set environment variables
 ENV PYTHONDONTWRITEBYTECODE 1
 ENV PYTHONUNBUFFERED 1
 
+ARG DJANGO_DEBUG=False
```
sboisvert, line 15: Why are we putting them here and not inside the ENV vars for Heroku?
Mostly curious because I'm currently having trouble getting env vars to show up where I want them so want to see if it's related

pcraig3: `ARG` means that it's an argument you can pass to the container when you build it. So we accept an arg called `DJANGO_DEBUG` and it defaults to false.

Keeping them in Heroku means we're platform dependent rather than having everything saved in the codebase, so it's better to put them somewhere in the code if possible. 

Some of the confusion though (I feel the same) is that the Heroku build is using `app.json` and then `heroku.yml`, so I wasn't clear what gets set when.

pcraig3: The thing I found is that setting it in the Heroku console seems to override the other ones, but setting env vars in app.json or heroku.yml wasn't working unless these `ARG`s existed in the container.

portal/settings.py
```diff
@@ -29,8 +29,9 @@
 # See https://docs.djangoproject.com/en/3.0/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = os.getenv(
-    "DJANGO_SECRET_KEY", "j$e+wzxdum=!k$bwxpgt!$(@6!iasecid^tqnijx@r@o-6x%6d"
+SECRET_KEY = (
+    os.getenv("DJANGO_SECRET_KEY")
+    or "j$e+wzxdum=!k$bwxpgt!$(@6!iasecid^tqnijx@r@o-6x%6d"
```
PierrePaul, line 34: neat hack

PR #52, opened 2020-06-29
Refactor SignupForm and SignupView

pcraig3:
This PR redoes the SignupForm to extend from the `UserCreationForm` and it uses a class-based view for the signup route. We don't have to merge this as-is (the code is already working without this), although I would push for the form refactor to go in because it removes a lot of password-based complexity that we don't have to think about.

Also, I found this really good blog post on `FormView`, which the difference between the functional view and the class view: https://www.agiliq.com/blog/2019/01/django-formview/

### Signup form

As I mentioned in another PR, the [`UserCreationForm`](https://github.com/django/django/blob/ec5aa2161d8015a3fe57dcbbfe14200cd18f0a16/django/contrib/auth/forms.py#L82) has a lot of password management code already in it, so there was a lot of copied and pasted code in our Form. Saving the user and checking the passwords happens in the UserCreationForm, so what we need to do is:

- Add the Validator to the `name` field
- Make sure `email` is 'readonly'
- `clean` the email by making sure it's one of the invited users

 ### Signup view as a class

The `SignupView` extends from `FormView`, so it handles most of the form stuff (ie, the 'get' and 'post' boilerplate, and form validation). We have to set the initial email value and then in `form_valid`, we save the user and return a success message.

### DIFF
--- a/profiles/forms.py
+++ b/profiles/forms.py
@@ -2,6 +2,7 @@ from django.contrib.auth.forms import (
     UserChangeForm,
     AuthenticationForm,
     PasswordResetForm,
+    UserCreationForm,
 )
 from django import forms
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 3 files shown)


profiles/forms.py
```diff
@@ -56,54 +57,28 @@ def __init__(self, *args, **kwargs):
...
-        self.cleaned_data = super().clean()
+
+    class Meta:
+        model = HealthcareUser
+        fields = ("email", "name")
+        widgets = {
+            "email": forms.EmailInput(attrs={"readonly": "readonly"}),
+        }
+
+    def clean_email(self):
```
sboisvert, line 73: Does this function run? 
I had to do it inside clean() instead of clean_email() because when making the field disabled it would ignore the actual data in the field and just keep the data in initial and not actually run the clean_email function

pcraig3: Yup, it runs no problem. Both error cases raise the validation messages we're expecting. Didn't know that's why you were doing `clean()` instead of the more specific one, but that makes sense.

PR #54, opened 2020-06-30
Local development setting enhancements

bryan-robitaille:
A few enhancements to facilitate local development.
- Refactored DEBUG variable to align with DJANGO_ENV so that a development environment will always have debug enabled.  Removed the need to have DJANGO_DEBUG explicitly declared.
- Refactored database settings so if a developer doesn't declare a database the project will use a default SQLite DB for local development.  Logic so a developer can still pass either a DB URL string or a DB declaration through env variables has not changed.
- Updated ReadMe in both languages with a quick start that leverages the entrypoint.sh as a way to save a local dev time in going through all the commands to launch an instance.

**This PR:**
- [x]  does not break any existing Tests
- [x]  has updated the README as required in both languages
- [x]  clearly identifies the changes made


Dockerfile
```diff
@@ -12,9 +12,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 ENV PYTHONDONTWRITEBYTECODE 1
 ENV PYTHONUNBUFFERED 1
 
-ARG DJANGO_DEBUG=False
-ENV DJANGO_DEBUG=$DJANGO_DEBUG
-
```
pcraig3, line 17: The thinking here was that we might sometimes want to run the app in debug mode when it's online so that we can see the errors we're getting — for example on our pull request review apps, the app runs in "debug" mode.

It does make our app more complex, but it's been useful so far. Not opposed to changing it, but wondering if that changes your mind.

portal/settings.py
```diff
@@ -94,14 +92,9 @@
...
-    db_config = {
-        "ENGINE": "django.db.backends.sqlite3",
-        "NAME": os.path.join(BASE_DIR, "db.sqlite3"),
-    }
-elif os.getenv("DATABASE_URL"):
-    db_config = dj_database_url.config(conn_max_age=600, ssl_require=True)
-else:
+default_db_path = os.path.join(BASE_DIR, 'db.sqlite3')
+
+if os.getenv("DATABASE_HOST"):
```
pcraig3, line 96: So this change means that in dev + tests we would use Postgres vs being able to do one for testing and one for development. 

In theory, that should be okay though. The original reason I put in the "TESTING" flag was so that our tests would pass on CI, because on Github actions we don't have a PG instance.

bryan-robitaille: Dev + tests will still use SQLite as no DATABASE_HOST or DATABASE_URL are being passed.  I just rearranged the logic a bit so that it checks first to see if DATABASE_HOST exists and if so to use those env variables.  If DATABASE_HOST does not exist it then checks for the DATABASE_URL env variable and if it's not declared uses a default SQLite DB locally.
This give the Dev env option of either using the Docker or local PostgreSQL options or just spin it up and go with a SQLite DB.

bryan-robitaille: I'm noting that it doesn't work well if a .env file is declared.  Will fix and push to this PR.

bryan-robitaille: Fixed by adding logic to detect if unset .env variables are used and to remove them from the env mapping to not interfere with app logic

Dockerfile
bryan-robitaille: Look forward to learning more about how the PR request review apps are setup but could they not also be set to DJANGO_ENV = "development"?

PierrePaul: Personally I set `DJANGO_DEBUG=True` directly in my `docker-compose.yml` file.
I assumed the docker-compose.yml was not read for production use.
Running `DEBUG` in production is a big security concern, so I like defaulting to False.

We will have to think about ALLOWED_HOSTS as well for production at some point.

bryan-robitaille:
Sorry for the late addition of the failsafe commit.  Was coming across this issue when doing auto-builds in AWS CodeBuilder.

PR #58, opened 2020-07-03
AWS configuration and deployment files

bryan-robitaille:
It may only look like three files but it took 30 commits of fine tuning to get them to what they are.  
These files will :
- Modify the Elastic Beanstalk NGINX config to change the host on the incoming request when it's from the ELB Health Checker.  This will ensure that the cluster health check is actually checking that the App is running and not just Nginx. 
- Build specs for the Continuous Delivery portion of Code Pipeline.  Builds the App, runs the tests and if everything passes it deploys.
- Database Migration script is run after app deployment but before the instance is accepting traffic.


.platform/hooks/postdeploy/00_migrate_db.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+source /var/app/venv/*/bin/activate
+echo "Migrate DB"
+python manage.py migrate --noinput
```
pcraig3, line 4: We probably don't want to migrate the DB every time on the prod instance. Is there a way to make this a manual trigger?

pcraig3: We can say it's staging and then all good though.

buildspec.yml
```diff
@@ -0,0 +1,33 @@
+version: 0.2
+
+env:
+  variables:
+    DJANGO_ENV: "development"
```
pcraig3, line 5: Is this what we want? It means we're running over http, not https, among other things.

buildspec.yml
```diff
@@ -0,0 +1,33 @@
...
+phases:
+  install:
+    runtime-versions:
+      python: 3.x
+  pre_build:
+    commands:
+      - pip install --upgrade pip setuptools pipenv
+      - pipenv install --dev
```
pcraig3, line 13: Do we need "dev" here? I think we only want prod stuff.

buildspec.yml
```diff
@@ -0,0 +1,33 @@
...
+      - pipenv run python manage.py sass profiles/static/scss/styles.scss profiles/static/css/
+      - echo "Collect static files"
+      - pipenv run python manage.py collectstatic --noinput -i scss
+      - echo "Compiling French Translations"
+      - pipenv run python manage.py compilemessages
+      - echo "Running DB Migrations locally"
+      - pipenv run python manage.py migrate --noinput
+  build:
+    commands:
+      - pipenv run python manage.py test
```
pcraig3, line 24: Why are we running tests as part of the build? Shouldn't GitHub actions be enough?

buildspec.yml
```diff
@@ -0,0 +1,33 @@
...
+
+  post_build:
+    commands:
+      - echo "Cleaning up"
+      - rm db.sqlite3
+      - rm ./portal/requirements.txt
```
pcraig3, line 30: We're using PG, right? The db is created as part of running the `test` command?

buildspec.yml
```diff
@@ -0,0 +1,33 @@
...
+      - pip install --upgrade pip setuptools pipenv
+      - pipenv install --dev
+      - echo "Compile CSS"
+      - pipenv run python manage.py sass profiles/static/scss/styles.scss profiles/static/css/
+      - echo "Collect static files"
+      - pipenv run python manage.py collectstatic --noinput -i scss
+      - echo "Compiling French Translations"
+      - pipenv run python manage.py compilemessages
+      - echo "Running DB Migrations locally"
+      - pipenv run python manage.py migrate --noinput
```
pcraig3, line 21: What's the difference between this migrate command vs the one above?

pcraig3: Wait this one is used to set up the tests, right? And then we run the tests and `rm` the database file? 

I think we should skip the tests here because Github Actions runs them for us.

buildspec.yml
bryan-robitaille: So this is only in the sandbox of the Build env which doesn't have internet access.  The AWS Codeline process is GitHub pull into Build sandbox.  All the commands in the Build sandbox are run.  Then only the files are copied out of the Sandbox and into the Elastic Beanstalk instance where the Django app uses the Environment Variables in the EB instance for the Django app.

buildspec.yml
bryan-robitaille: You'll want dev here if we want to add a test again for format, ect.  None of these packages get carried over into the production hosting env.

buildspec.yml
bryan-robitaille: Because a merge can be done on the main branch outside of github there are ways to bypass Github actions.  Accidental or not this just provides an extra security blanket to make sure no accidents do make it through to prod or staging.

buildspec.yml
bryan-robitaille: In the Elastic Beanstalk env it does connect to PG.  In the build sandbox it doesn't have access to anything outside of that Ubuntu instance which is why it uses SQLite locally.  To make sure the db.sqlite3 file isn't copied over to the Elastic Beanstalk instance it's deleted before the files (aka artifacts) are copied over.

.platform/hooks/postdeploy/00_migrate_db.sh
bryan-robitaille: If there are no migrations to be run then command returns "No migrations to apply".  The danger of not running the migrations on every deployment is that if a major feature change, like the province change to the models, and the deployment goes through without the migration then the Django app breaks.  There are few things I have in mind in regard to Staging vd Prod that we can go through next week.  Essentially Continuous Deployment is only setup on Staging and once we're happy with a version we can 'deploy it' to the Production environment with a click of a button.

pcraig3: Fair enough, sounds right to me.

buildspec.yml
pcraig3: Yup, got it.

buildspec.yml
pcraig3: > None of these packages get carried over into the production hosting env.

Right, so we're not building and running the prod instance in the same step.

buildspec.yml
bryan-robitaille: Running these tests again just covers our bases in case commits are pushed directly to main and bypasses a PR.

PR #59, opened 2020-07-06
Add in "manage accounts" + invite users + user account page + removing users

pcraig3:
This PR does lots of things, but looks more confusing because it includes the code for the "Provinces" model because that is still open: #57. If we merge that, I can rebase this and the commit history is shorter.

- Allow us to create admin users from the admin interface
- Fixed a few a11y issues that were getting flagged by my `axe` scanner
- Added `is_superuser` attribute to our user profiles (before we just had regular users and admins)
- Added the "invite" view so that admins can invite users without the admin interface
- Added the "manage accounts" page where admins can see accounts for all users from their province (including other admins)
- Added the "profile" page (and a link at the top) so that accounts can see their own info
- Added a "delete" view so admins can remove other accounts
- Refactored the tests so there is less copy+paste

Honestly, I think we should just lock non-superusers out of the admin interface because you can do whatever you want in there.

At this point, I think functionality-wise, we're basically at MVP.

## Screenshots

| description |  screen  |
|---|---|
| "start" page with "manage accounts" link in the main nav and "account" link in the top nav | Screen Shot 2020-07-06 at 9 44 07 AM |
| "manage accounts" page for admins. admins can see all other users from their province.  | Screen Shot 2020-07-06 at 9 44 19 AM |
| "user profile" page. all users can see their own profiles, and admins can see profiles of users from their province.  | Screen Shot 2020-07-06 at 9 44 11 AM |
| "invite" page. admins can invite new users by email.  | Screen Shot 2020-07-06 at 9 44 49 AM |
| "invite confirm" page. after an account is invited, you see the confirmation page  | Screen Shot 2020-07-06 at 9 44 39 AM  |
|  "remove an account" page. admins can remove accounts from their province (except themselves)  |  Screen Shot 2020-07-06 at 9 45 01 AM |


sboisvert:
>Honestly, I think we should just lock non-superusers out of the admin interface because you can do whatever you want in there.
👍 
>At this point, I think functionality-wise, we're basically at MVP.
🚀

sboisvert:
Can admins remove other admins?

profiles/templates/invitations/templates/invite_complete.html
```diff
@@ -0,0 +1,22 @@
...
+{% block title %}
+  {% if form.errors %}{% trans "Error: "%}{% endif %}
+  {% trans "Invitiation sent"  %}
+{% endblock %}
+
+{% block content %}
+  <div class="jumbotron">
+    <h1>{% trans "Invitation sent" %}</h1>
+
+    {% if request.session.invite_email %}
```
sboisvert, line 13: Is there a case where this wouldn't be true? 
If fine if it's defensive programming, just want to know for personal knowledge.

profiles/templates/profiles/healthcareuser_list.html
```diff
@@ -1,18 +1,38 @@
...
+          <th scope="col">{% trans "Actions" %}</th>
+        </tr>
+    </thead>
+    <tbody>
+      {% for healthcareuser in healthcareuser_list %}
+        <tr>
+          <td scope="col"><a href="{% url 'user_profile' healthcareuser.id %}">{{ healthcareuser.email }}</a></td>
+          <td scope="col">{{ healthcareuser.is_admin|yesno:_("Admin,Staff") }}</td>
+          <td scope="col">
+          {% if not healthcareuser.is_admin %}
```
sboisvert, line 25: This answers my question of can they delete other admins.
I guess we just need to setup a process for them to ask us to delete admins for the MVP

profiles/urls.py
```diff
@@ -18,14 +19,32 @@
         login_required(TemplateView.as_view(template_name="profiles/start.html")),
         name="start",
     ),
+    path(
+        "invite/",
+        staff_member_required(views.InviteView.as_view(), login_url="login"),
```
sboisvert, line 24: Why do we need to pass login_url?

profiles/views.py
```diff
@@ -57,6 +58,86 @@ def form_valid(self, form):
...
 
 
+class InviteView(FormView):
+    form_class = HealthcareInviteForm
+    template_name = "invitations/templates/invite.html"
+    success_url = reverse_lazy("invite_complete")
+
+    def get_initial(self):
+        # preload the invite form with the id of the inviter
+        return {"inviter": self.request.user.id}
```
sboisvert, line 68: I guess this also answers my question about the inviter in the invitation flow.

profiles/views.py
```diff
@@ -57,6 +58,86 @@ def form_valid(self, form):
...
+        invite = form.save()
+        invite.send_invitation(self.request)
+        messages.success(
+            self.request, "Invitation sent to “{}”".format(invite.email), "invite"
+        )
+        self.request.session["invite_email"] = invite.email
+        return super().form_valid(form)
+
+
+class ProfilesView(LoginRequiredMixin, UserPassesTestMixin, ListView):
```
sboisvert, line 81: This is really sweet functionality

profiles/views.py
```diff
@@ -57,6 +58,86 @@ def form_valid(self, form):
...
+
+class UserProfileView(LoginRequiredMixin, ProvinceAdminManageMixin, View):
+    def get(self, request, pk):
+        profile_user = get_object_or_404(HealthcareUser, pk=pk)
+        return render(
+            request, "profiles/user_profile.html", {"profile_user": profile_user}
+        )
+
+
+class ProvinceAdminDeleteMixin(ProvinceAdminManageMixin):
```
sboisvert, line 126: So admins could delete other admins, the only thing stopping them is specifying the right URL? (since it wouldn't be linked in the view)

profiles/forms.py
```diff
@@ -100,3 +102,28 @@ class Meta:
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Otherwise it just says "Email"
+        self.fields["email"].label = _("Email address")
+
+    def clean_inviter(self):
+        # get the id and return the instance
+        user_id = self.cleaned_data.get("inviter")
```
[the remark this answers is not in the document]
pcraig3: No, that's a really good point actually. We should just use the logged-in user when creating the invite.

profiles/migrations/0006_healthcareuser_is_superuser.py
```diff
@@ -0,0 +1,18 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('profiles', '0005_add_provinces'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='healthcareuser',
+            name='is_superuser',
```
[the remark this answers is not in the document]
pcraig3: Yeah, I think it's fine here. We don't want 2 different classes of users, we just want 1 with different permissions. The `HealthcareUser` is named that because I subclassed the default `User` but didn't want to reuse the name.

Superusers don't really need a province, for example, but if we had 2 models floating around, then we would introduce a string of if/else checks through the app that wouldn't really be useful.

profiles/static/js/script.js
```diff
@@ -1,12 +1,21 @@
...
-for (var i = 0, len = elements.length; i < len; i++) {
-  elements[i].addEventListener('keydown', function (e) {
+for (var i = 0, len = buttonLinks.length; i < len; i++) {
+  buttonLinks[i].addEventListener("keydown", function (e) {
     if (e.keyCode == 32) {
       e.target.click();
       e.preventDefault();
     }
   });
+} // Add disabled to buttons after click
```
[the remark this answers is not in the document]
pcraig3: Yeah, good point, I can move it.

profiles/static/js/script.js
```diff
@@ -1,12 +1,21 @@
...
+
+
+var form = document.querySelector("main form");
+
+if (form) {
+  form.addEventListener("submit", function (e) {
+    form.querySelector("button").disabled = true;
```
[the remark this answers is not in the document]
pcraig3: The reason I put this in is because sending the email on the "invite" page takes ~3-5 seconds without any visible loading indicator, and if you press the "submit" button again during that time it will screw up your invite. 

Refreshing the page would reenable the button, as would getting a form error or hitting the "back" button in the browser. I can't think of a scenario where this would lock someone out, so I think it's safe.

We _could_ put in a 5 second timeout or something, just to be safe, but I would like to hear a plausible scenario of it going wrong before adding extra logic.

profiles/templates/includes/nav_main.html
```diff
@@ -1,10 +1,13 @@
 {% load i18n %}
 
-<nav class="nav--main">
+<nav class="nav--main"  aria-label="{% trans 'Product navigation' %}">
   <div class="page--container">
     <ul>
-      <li><a class="{% if request.resolver_match.url_name == 'start' %}active{% endif %}" href="{% url 'start' %}">{% trans "Start" %}</a></li>
-      <li><a class="{% if request.resolver_match.url_name == 'code' %}active{% endif %}" href="{% url 'code' %}">{% trans "Get a number" %}</a></li>
+      <li><a {% if request.resolver_match.url_name == 'start' %}class="active"{% endif %} href="{% url 'start' %}">{% trans "Start" %}</a></li>
```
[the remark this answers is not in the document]
pcraig3: I agree this is crazy, but I can't think of one. I looked this up on [StackOverflow](https://stackoverflow.com/questions/25044370/make-clicked-tab-active-in-bootstrap) and saw a lot of versions of this.

profiles/templates/invitations/templates/invite_complete.html
pcraig3: Yeah, you can visit this page manually as an admin. So if you log in and then just type in the URL you can get to this page without sending an invite. It feels like someone who is doing that probably knows what they are doing, so we don't need to have a message being like "hey, this didn't really send an invite". But we could also redirect or something if it's not in the session.

profiles/templates/profiles/healthcareuser_list.html
pcraig3: They actually can delete admins, they just can't see the link here.

If they manually type in the URL, they can delete other admins from the same province. Interestingly, they would be able to delete superusers from the same province too, so I'll update this.

On Marketplace, we initially had a complicated admin process because we didn't want one admin to be able to go rogue and delete all the other admins and every other account. But then since it was hard for users to self-administer, a lot of requests were going to our support person and she said we should let admins delete other admins. So we did and it was fine.

profiles/urls.py
pcraig3: Otherwise it takes us to the `admin:login` route. To be honest, I think I should refactor this to use UserMixins because that's how I did it in the other views.

profiles/views.py
pcraig3: No, you're right from before. Instead of passing the variable through the form, we can get it on the server when the POST request comes back.

profiles/views.py
pcraig3: Yes, exactly. See https://github.com/cds-snc/covid-healthcare-portal/pull/59#discussion_r450897824

profiles/templates/profiles/healthcareuser_list.html
sboisvert: Ya, I think we can let admins delete other admins but not super users. Might as well make the link visible and not make them guess. At least with the link visible it's clear that it's an intended behaviour that we're allowing

PR #62, opened 2020-07-06
Super User auto configuration + Settings refactor

bryan-robitaille:
A few things added here:

1. Remove hardcoded 'elasticbeanstalk.com' and 'heroku' from ALLOWED_HOSTS.  Instead append to the list that is used for local dev by listening for an environment variable ALLOWED_HOSTS.  Multiple entries can be submitted in a comma separated list.  I've added the 'heroku' entry as a var in the Dockerfile because the Heroku config was using the docker image for hosting.

2. Create a default super user... only if you want to.  When spinning up new deployments it's not always evident on trying to access the virtual environment to run the createsuperuser command.  This custom admin command allows the creation of a super user through the cli by using environment variables.

**It:**
- Checks to see if the flag is set to create a default super user `DJANGO_DEFAULT_SU`
- Checks to see if there is already an existing user with that name.  If so do nothing.
- Ensures that `DEFAULT_SU_PASSWORD` is set and validates it against the complexity requirements set in `portal.settings`
- If if doesn't meet the complexity requirements it throws a validation error.
- If it meets the validation requirements it creates the new super user.


Dockerfile
```diff
@@ -15,6 +15,8 @@ ENV PYTHONUNBUFFERED 1
 ARG DJANGO_ENV=production
 ENV DJANGO_ENV=$DJANGO_ENV
 
+ARG ALLOWED_HOSTS=.herokuapp.com
```
pcraig3, line 18: This will get overwritten otherwise right? So on Heroku we don't set it but in other places we do?

profiles/management/commands/createdefaultsuperuser.py
```diff
@@ -0,0 +1,29 @@
...
+
+SU_EXISTS = HealthcareUser.objects.filter(name="hcw_portal_admin").exists()
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        if CREATE_DEFAULT_SU and SU_DEFAULT_PASSWORD and not SU_EXISTS:
+            self.create_su()
+        elif SU_EXISTS:
+            self.stdout.write(self.style.SUCCESS("Default Super User already exists"))
```
pcraig3, line 19: This should be the first `if`, otherwise we will still try to create one.

profiles/management/commands/createdefaultsuperuser.py
```diff
@@ -0,0 +1,29 @@
...
+        else:
+            self.stdout.write(self.style.SUCCESS("Not creating default Super User"))
+
+    def create_su(self):
+        ## Check to ensure SU password meets minimum requirements
+        if validate_password(SU_DEFAULT_PASSWORD) is None:
```
pcraig3, line 25: There's no message for users if the password doesn't work. It seems like that would make it fail silently.

Dockerfile
bryan-robitaille: Put this in place so the current Heroku configuration wouldn't need to be modified.  Only place this Allowed_Host would be used is in that docker image which is currently being used for Heroku and local dev.  Would prefer to move that variable to the docker-compose file but I'm not sure if you can set runtime variables in Heroku for docker images.  Ideally the Allowed_Hosts would be set at runtime instead of during the Docker image build.

profiles/management/commands/createdefaultsuperuser.py
bryan-robitaille: The SU_EXISTS check is part of the first `if`.
`if CREATE_DEFAULT_SU and SU_DEFAULT_PASSWORD and not SU_EXISTS`

profiles/management/commands/createdefaultsuperuser.py
bryan-robitaille: The Validation Error returned by the validate_password is raised and displays an error message in stdout with why the password didn't pass validation.  I could get fancy and catch the errors and then echo them out.  Would still need to throw a validation error though otherwise the test/build will think it passed.

Dockerfile
sboisvert: Should we just make it blank and always inject it with the env vars?

profiles/management/commands/createdefaultsuperuser.py
pcraig3: Ah, all good then.

bryan-robitaille:
> Looks great!
> I wonder if we should have a check that disables this if we're running on production.
> It feels a bit janky to be relying on a script that creates a SU on production (and also potentially error prone / gives out potential username information for a production system).
> The added benefit of not having it running for production is we could then just skip the password validation requirements since it's not on any important system.

It doesn't need to run on production.  We just don't include the DJANGO_DEFAULT_SU env variable and it's like the script doesn't exist.  You could also just set the env variable to True for the first run and once created remove the env variable from the cluster.  By doing so you don't need to modify the security group to open port 22 on the Virtual machine.

profiles/management/commands/createdefaultsuperuser.py
bryan-robitaille: Reread your comment, I think I missed understood what you meant.  I see what you mean.  Will re look at the logic tomorrow morning with fresh eyes.

bryan-robitaille:
Removed unused imports and refactored logic as suggested by @pcraig3

profiles/management/commands/createdefaultsuperuser.py
```diff
@@ -0,0 +1,27 @@
...
+from profiles.models import HealthcareUser
+
+SU_EXISTS = HealthcareUser.objects.filter(name="hcw_portal_admin").exists()
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        if SU_EXISTS:
+            self.stdout.write(self.style.SUCCESS("Default Super User already exists"))
+        elif CREATE_DEFAULT_SU and SU_DEFAULT_PASSWORD and not SU_EXISTS:
```
pcraig3, line 13: Since we already check for `SU_EXISTS`, we don't need to check the `not` case here. This can just be 
```
 elif CREATE_DEFAULT_SU and SU_DEFAULT_PASSWORD:
```

PR #63, opened 2020-07-07
Add config require for New Relic monitoring

bryan-robitaille:
Add the import required in pipfile and wsgi.py file.
In AWS the New Relic config will be set with the following env variables through Elastic Beanstalk.
NEW_RELIC_LICENSE_KEY
NEW_RELIC_APP_NAME


portal/wsgi.py
```diff
@@ -6,11 +6,15 @@
...
 """
-
+import newrelic.agent
 import os
-
 from django.core.wsgi import get_wsgi_application
 
+newrelic.agent.initialize()
+# Uses the following env vars to set config
+# NEW_RELIC_LICENSE_KEY
```
sboisvert, line 15: I'm not sure if it's useful but maybe we should add it to the .env.example file so we have one place where we have all of them listed?

bryan-robitaille: That's a good idea.  I also noticed that the ReadMe probably also probably needs an update with all these new env vars over the last couple of PR's.  Will open a new PR where I'll go through the project and ensure all the env vars are listed in the .env.example as well as a listing in the ReadMe on what they control.  Was also thinking that in the near future we'll need to look at which way we want to lean for the env var defaults.  Do we lean towards Dev or Production for the defaults.

PR #64, opened 2020-07-07
lock healthcare administrators out of /admin/

sboisvert:
We will be giving the is_admin property to healthcare administrators and using is_superuser for CDS folks.


pcraig3:
Closing because this went in in #65

PR #68, opened 2020-07-08
Add @PierrePaul to CODEOWNERS

pcraig3:
Means that PierrePaul can approve PRs just like others can.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,2 @@
 # CDS developers must review all changes to ensure security of changes
-*   @cds-snc/exposure-notification-devs
+*   @cds-snc/exposure-notification-devs @PierrePaul


sboisvert:
@pcraig3 I would double check with @jeffmaher on this. I think the idea was that only you and me as Full time CDS members should be able to approve as part of the code owners group. 
I think we want @bryan-robitaille and @PierrePaul to give reviews and to approve and reject commits, but I think we want to keep having it "forced" that a full time CDS person also gives the approval.

sboisvert:
I just realized @bryan-robitaille has CODEOWNERS approval authority. Maybe I'm wrong in interpreting what @jeffmaher mentioned, either way, it would be good to be consistent :)

jeffmaher:
@bryan-robitaille does have approval authority (since he's from GC before this), but not @PierrePaul. I realize this isn't desirable or convenient, and I hear you. Alas, it's a condition of our rapid security clearance process to get fast hires onboard quickly. This ensures that someone from CDS/GC that has gone through the longer security review process has seen+approved any code submissions.

jeffmaher:
Still great to get @PierrePaul 's review comments and approve/not approves, just with the caveat that if approve, someone at CDS also needs to do a quick check to approve.

PR #72, opened 2020-07-08
Limit SESSION_COOKIE_AGE to 20 hours

sboisvert:
Related to SR-4 Security we'll start by limiting the session to 20h. This should mean that folks only need to login once a day, probably in their morning.

If user research demonstrates that this is too burdensome we can revisit it later.


sboisvert:
[Auth'd users' sessions time out after inactivity](https://trello.com/c/V5PexF8G/46-authd-users-sessions-time-out-after-inactivity)

PR #73, opened 2020-07-08
Add Axes Login Limiter

sboisvert:
This is a draft pull request that needs more testing.

We need to lockout users after a certain amount of login attempts.

https://trello.com/c/MipISjdj/107-add-login-limitations

https://django-axes.readthedocs.io/en/latest/index.html


PierrePaul:
I'm not sure we are going to be able to add unit testing to Axes.
It requires a valid request object to be sent to `authenticate` (which is used by `this.client.login`) and our request is faked by the test suite.

Official axe documentation just says to disable axe when running automated testing. I have updated the settings file and the Pipfile to be able to do that.

PierrePaul:
Also side note, right now it gives us a white page with this message.

Screenshot_2020-07-09 Screenshot

PierrePaul:
I added a single test that check at least for super user to be locked out after 10 tries (taken from settings.py).
It's the only test that runs with AXES enabled.
@sboisvert I think it's safe to remove the draft flag on this PR.

bryan-robitaille:
Up for discussion but I see 2 potential concerns:
- A 1hr cool off is a really long time.  In the past in other apps we've set this at 5 min. (GCaccount)
- The `AXES_ONLY_USER_FAILURES = False` may also raise issues with health organizations that sit behind NATs.  We disabled this in GCaccount because if a single user behind a NAT locks out their account every user that leverages the same IP behind that NAT will also not be able to connect.

sboisvert:
After digging a bit more into this I found that it's not going to be super straight forward to get an actual HTML page up. Right now it just outputs the value of COOLOFF_MESSAGE [conf.py](https://github.com/jazzband/django-axes/blob/3ceb546ae215b183b3d28e64ec1fd15abc1f036d/axes/conf.py#L79)
There's a way to [create a callable](https://github.com/jazzband/django-axes/blob/3ceb546ae215b183b3d28e64ec1fd15abc1f036d/axes/conf.py#L48) that we can probably use to generate a template.
Unless there are concerns from others I'm going to punt having a full template to a future version. I don't think it brings enough value for us to prioritize it.

See https://trello.com/c/MipISjdj/107-add-login-limitations

sboisvert:
@bryan-robitaille This should address your concerns. Let me know if it works for you.

bryan-robitaille:
This is how we did it in Concierge if ever you were looking for a way to achieve this: https://github.com/gctools-outilsgc/concierge/pull/118

portal/settings.py
```diff
@@ -209,3 +218,19 @@
 # Default Super User Setup
 CREATE_DEFAULT_SU = os.getenv("DJANGO_DEFAULT_SU", "False") == "True"
 SU_DEFAULT_PASSWORD = os.getenv("SU_DEFAULT_PASSWORD", None)
+
+# Login Rate Limiting
+if os.getenv("DJANGO_ENV", "production") == "tests":
+    AXES_ENABLED = False
+
+AXES_FAILURE_LIMIT = 5  # Lockout after 5 failed login attempts
+AXES_COOLOFF_MESSAGE = _(
```
bryan-robitaille, line 227: Missing french translation in PO file

PR #75, opened 2020-07-08
Update content + add french translations 

pcraig3:
This PR updates the content and adds french translations for the `Archived (0.2)` version of the app in [Figma](https://www.figma.com/file/znPqLTQQDcUkaNjdDcShiz/Exposure-Notification---Health-Portal-(mock-up)?node-id=918%3A0) (the current version is still being worked on).

It doesn't add any new pages but this is most of what we need and it's standalone so I figured we could merge this first.

Pages included in this PR:

- Landing page
- Login page
- Password Reset page
- Password Reset sent page
- "Start" page, intro to the portal
- "Code" page
- User profile page
- "Manage accounts" page
- Invite a user page

Still to be done:

- Add a privacy page
- Signup page
- Instructions page (is that part of the MVP?)
- Buttons for forms aren't translating
- "Password reset successfully" pages


profiles/tests.py
```diff
@@ -124,7 +124,7 @@ def test_loginpage(self):
         #  Get the login page
         response = self.client.get(reverse("login"))
         self.assertEqual(response.status_code, 200)
-        self.assertContains(response, "Login")
+        self.assertContains(response, "<h1>Log in</h1>")
```
sboisvert, line 127: I think we should at some point add some tests that check translations. Might need not having static text strings here.

pcraig3: We have some tests that check for the landing page with different language based on the browser cookie or the url string.

PR #76, opened 2020-07-09
Feature user timeout

PierrePaul:
Limiting session time to 5 minutes or when closing the browser (Chrome ignores the setting, to keep in mind). 
Using django-session-security, now showing a warning at 4 minutes that the session is going to expire.

Screenshot_2020-07-08 Generate a number

**Important note** : django-session-security requires jquery 2.x. I have added it as a static file. Let me know if you would prefer a CDN (and which one please).


sboisvert:
I set the session timeout to be 20h which I think is good enough. In a perfect world I'd rather not add jQuery just to get a activity timeout going. I also don't think it's required for the security of the app. having folks logged out once a day should be good enough based on how we think people will use the app. Thoughts @pcraig3 ?

profiles/templates/base.html
```diff
@@ -48,5 +48,7 @@
 
     </div>
     <script src="{% static 'js/script.js' %}"></script>
+    <script src="{% static 'js/jquery-2.2.4.min.js' %}"></script>
+    {% include 'session_security/all.html' %}
```
pcraig3, line 52: Do we need the JS popup? Otherwise we could do this purely server side and not worry about jQuery.

pcraig3: ie, something like this which is server-centric might fit the bill: https://pypi.org/project/django-session-timeout/

pcraig3:
> I set the session timeout to be 20h which I think is good enough. In a perfect world I'd rather not add jQuery just to get a activity timeout going. I also don't think it's required for the security of the app. having folks logged out once a day should be good enough based on how we think people will use the app. Thoughts @pcraig3 ?

I agree. 

- I think extend the session time: 20 hours seems fine
- Having a popup every 5 minutes is going to stop people from using the app. If we ditch the popup, we don't need the JQuery dependency

bryan-robitaille:
At GCTools for Concierge (GCaccount single sign on) we used 24 hours as our session timeout.  Agree completely that forcing people to log in too many times during a work day will lower user satisfaction with the app

obrien-j:
Re: Timeouts. This is definitely a security issue, but as long as you have a timeout configured, I'm content.

20hr might be a bit much, but I would be operating on the assumption that their local machines with have a much shorter screen/lockout functionality. If a session gets leaked, whether it's 5 minutes or 20 hours, we're going to see a massive spike of key requests, so it might be worth putting more effort into rate limiting the key generation on the portal itself, to something that is reasonable for a human.

sboisvert:
I didn't create a clear enough card for this and I caused confusion, It was done in #72

PR #77, opened 2020-07-09
Now generating the expiry via javascript and showing local time

PierrePaul:
Now generating the expiry time in javascript and adding 24 h. Also displaying the local time in both french and english format.

Screenshot_2020-07-09 Fournir le numéro au patient

----

Screenshot_2020-07-09 Give patient this number

### DIFF
--- a/profiles/templates/profiles/code.html
+++ b/profiles/templates/profiles/code.html
@@ -8,19 +8,7 @@
   <h1>{% trans "Give patient this number" %}</h1>
   <div id="big-code">
     <code>{{ code }}</code>
-
-    <span>
-    {% trans "Number expires" %}
-    {% timezone "Canada/Eastern" %}
-    {% get_current_language as LANGUAGE_CODE %}
-    {% if LANGUAGE_CODE == 'en' %}
-        {{ time|date:"F jS \a\t f a" }}
-    {% else %}
-        {{ time|date:"j F à G\hi," }}
-    {% endif %}
-      {% trans "local time" %}
-    {% endtimezone %}
-    </span>
+    {% include "includes/expires_string.html" %}
   </div>
 
   <p>{% trans "Tell patient to:" %}</p>
--- a/profiles/templates/includes/expires_string.html
+++ b/profiles/templates/includes/expires_string.html
@@ -1,0 +1,39 @@
+{% load i18n %}
+{% load tz %}
+{% get_current_language as LANGUAGE_CODE %}
+
+{% timezone "Canada/Eastern" %}
+<span>{% trans "Number expires" %}</span>
+  <span id="expiry-date">
+    {% if LANGUAGE_CODE == 'en' %}
+        {{ expiry|date:"F j" }}
+    {% else %}
+        {{ expiry|date:"j F" }}
+    {% endif %}
+  </span>
+  <span>{% trans "at" %}</span>
+  <span id="expiry-time">
+    {% if LANGUAGE_CODE == 'en' %}
+        {{ expiry|date:"f a e" }}
+    {% else %}
+        {{ expiry|date:"G \h i e" }}
+    {% endif %}
+  </span>
+{% endtimezone %}
+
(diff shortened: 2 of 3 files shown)


profiles/static/js/script.js
```diff
@@ -10,3 +10,8 @@ for (var i = 0, len = elements.length; i < len; i++) {
   });
 }
 
+Date.prototype.addHours = function(h) {
+  this.setTime(this.getTime() + (h * 60 * 60 * 1000))
+  return this;
+}
+
```
pcraig3, line 17: Just FYI, this is loaded at the bottom of the `<body>` tag, but since your code is waiting for "DOMContentLoaded", it seems like it works just fine.

profiles/templates/profiles/code.html
```diff
@@ -5,24 +5,14 @@
...
-    {% get_current_language as LANGUAGE_CODE %}
-    {% if LANGUAGE_CODE == 'en' %}
-        {{ time|date:"F jS \a\t f a e" }}
-    {% else %}
-        {{ time|date:"j F à G\hi e" }}
-    {% endif %}
-    {% endtimezone %}
-    </span>
+    <span>{% trans "Expires" %}</span>
+    <span id="expiry"></span>
```
pcraig3, line 13: We prefer using [progressive enhancement](https://www.gov.uk/service-manual/technology/using-progressive-enhancement) at CDS so if anything goes wrong with the JS, it will still show the time.

I would say that instead of removing the server time, we could:

- overwrite it (ie, put it in "#expiry" and then replace the content
- put it in `<noscript>` tags so it shows up if JS is explicitly not available

We could also create a `.js-enabled` body class, but we don't really need that for this story.

profiles/templates/profiles/code.html
```diff
@@ -47,4 +37,10 @@ <h2>{% trans "Instructions for patient if needed" %}</h2>
   <p>{% trans "COVID Alert will ask permission to upload random codes daily for the next 13 days." %}</p>
 
   <div class="content-button"><a href="{% url 'start' %}" role='button' draggable='false'>{% trans "Next patient" %}</a></div>
+  <script type="text/javascript">
+      document.addEventListener('DOMContentLoaded', function(){
+          const expiry = new Date().addHours(24);
+          document.getElementById('expiry').innerHTML = expiry.toLocaleString('{{ LANGUAGE_CODE }}');
+      }, false);
+  </script>
```
pcraig3, line 45: Nice, this looks clean. 

We'd like to keep a similar format as we had before though (I know vanilla JS date formatting is awful), so I'm not sure what's the cheapest solution here.

Maybe [pulling in date-fns from an CDN](https://unpkg.com/)?

pcraig3:
Heyo, we want more control over the format of the "expires" string. 

What we have now in Figma is pretty different than we get here, so maybe bring in date-fns as a CDN dependency or maybe there's a better way to do this? I think we should say away from npm if it's only for one dependency. 

Screen Shot 2020-07-09 at 10 01 57 AM
Screen Shot 2020-07-09 at 10 01 46 AM

profiles/templates/profiles/code.html
PierrePaul: That was my first reflex, but the latest version of the (official) CDN for date-fns was 1.9 (4 years ago). unpkg does seem to have a good up to date version (2.14) so I will make the switch. Thanks!

profiles/static/js/script.js
PierrePaul: Thanks, this will probably go away when I switch to date-fns.

PierrePaul:
After talking with Paul, we had too many issues using date-fns without using NPM. So here is the updated pull request with the smallest version of momentjs, delivered by the cloudfare DNS.
The new format and their translations : 
French
Screenshot_2020-07-09 Fournir le numéro au patient(2)

-----

English
Screenshot_2020-07-09 Give patient this number(2)

PR #78, opened 2020-07-09
Translating button for submit text

PierrePaul:
Make submit_text translatable and update the translations for those buttons.

https://trello.com/c/l69tARDh/114-bug-form-buttons-are-not-translate-able-currently


docker-compose.yml
```diff
@@ -15,6 +15,7 @@ services:
       - DATABASE_PASSWORD=password
       - DATABASE_HOST=db
       - DATABASE_PORT=5432
+      - DJANGO_ENV=development
```
pcraig3, line 18: Is this intentional? (Not a problem, just wondering why)

PierrePaul: It is yes, if we dont put that in, settings.py thinks it should run the is_prod something. Which means it will redirect to https (which is not setup in our docker).

PR #79, opened 2020-07-09
Static files Configuration for Elastic Beanstalk

bryan-robitaille:
Adding configuration so that Elastic Beanstalk knows where to serve the static files from instead of relying on the Django app.


bryan-robitaille:
Not ready for review yet.  Going to look into the whitenoise package and see if it's still needed given our local dev process and our other hosting environments.

bryan-robitaille:
So this PR will work on Heroku, Docker, and AWS.
- Docker users the command `python manage.py runserver`
- Heroku now uses `uwsgi`with a path mapping to the static files
- AWS uses `aws:elasticbeanstalk:environment:proxy:staticfiles:` config to set nginx rules from /static

PierrePaul:
:+1:

Dockerfile
```diff
@@ -12,11 +12,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 ENV PYTHONDONTWRITEBYTECODE 1
 ENV PYTHONUNBUFFERED 1
 
-ARG DJANGO_ENV=production
-ENV DJANGO_ENV=$DJANGO_ENV
-
-ARG ALLOWED_HOSTS=.herokuapp.com
-ENV ALLOWED_HOSTS=$ALLOWED_HOSTS
```
pcraig3, line 19: Are we not using these anymore? We still want to keep review apps up and running. Will these be set elsewhere?

bryan-robitaille: Already set now in the env config on heroku

PR #81, opened 2020-07-09
Modify default SU email and Heroku Config

bryan-robitaille:
- Heroku config will now create the default super user during PR deployments
- Default SU admin email changed to admin@cds-snc.ca
- Refactored `createdefaultsuperuser` logic to remove duplicate logic check as per @pcraig3 suggestion.


bryan-robitaille:
Woot Woot it works on this PR's deployment

PR #83, opened 2020-07-09
Add translations for signup + remove visible province

pcraig3:
This PR does a few tweaks to the signup form:

- Updates the English and French content
- Makes the help text look more uniform
- Makes the province into a hidden field so that it's not visible to users signing up

And it also updates the "user profile" view:

- Uses `DetailView` instead of `View` for healthcare profiles
- Only admin users can see the "back to accounts" button

## Screenshots

| before | after   |
|---|---|
| Screen Shot 2020-07-09 at 6 16 32 PM | Screen Shot 2020-07-09 at 6 17 36 PM |


profiles/forms.py
```diff
@@ -61,13 +61,11 @@ class SignupForm(HealthcareBaseForm, UserCreationForm):
...
 
-    # disabled fields aren't submitted
+    # disabled fields aren't submitted / ie, can't be modified
     email = forms.CharField(
         widget=forms.TextInput, label=_("Email address"), disabled=True
     )
-    province = forms.CharField(
-        widget=forms.TextInput, label=_("Province"), disabled=True
-    )
+    province = forms.CharField(widget=forms.HiddenInput, disabled=True)
```
sboisvert, line 68: Do we even need to pass it on the form? 
If the data can't be modified because it's disabled, wouldn't it just be taking it from elsewhere (the session I'm assuming) anyway?

pcraig3: We definitely could, but this is easier for now.

PR #86, opened 2020-07-13
Admin can resend an invite link to a user

PierrePaul:
Deleting all valid non-accepted invitations for the same email by defaut, when creating an invitation


PierrePaul:
Imports cleaned up! Thanks

pcraig3:
Looks good 👍

PR #87, opened 2020-07-13
Language switcher across domains and development

bryan-robitaille:
Adds a simple template tag to handle language switching.
- Listens for a new env var `URL_DUAL_DOMAINS`.  Defaults to false.
- Configurable English and french dual domains through `URL_EN_PRODUCTION` and URL_FR_PRODUCTION`

Allows control over EN and FR production URL's from settings.py and default to current production URLs.


profiles/templatetags/language_switcher.py
```diff
@@ -0,0 +1,29 @@
...
+            '<a role="button" href="{}">{}</a>',
+            settings.URL_EN_PRODUCTION + translate_url(path, lang.get("code")),
+            lang.get("name_local"),
+        )
+    else:
+        return format_html(
+            "<a role='button' href='{}'>{}</a>",
+            settings.URL_FR_PRODUCTION + translate_url(path, lang.get("code")),
+            lang.get("name_local"),
+        )
```
pcraig3, line 29: Hello! I have a few observations about this so I'm just going to list them out.

- If this works, then can't we get rid of the web form altogether? 

- Let's use Title Case for the language string (ie, make the first character capitalised) 

- The "onclick" function won't run because of too many double quotes in the rendered HTML. You would have to do `onclick="document.forms[\"language_toggle\"].submit();"` for this to render properly.

- Most of the code in the render blocks is the same, so I think it would be better to isolate the logic around changing the href. eg,

```py
def get_href(context, lang):
    path = context.get("request").get_full_path()
    root_domain = ''
    lang = lang.get("code")

    if settings.URL_DUAL_DOMAINS:
        root_domain = settings.URL_FR_PRODUCTION if lang == "fr" else settings.URL_EN_PRODUCTION
        
    return root_domain + translate_url(path, lang)
```

bryan-robitaille:
> Using a form to switch languages is something I took from [the Django documentation](https://docs.djangoproject.com/en/3.0/topics/i18n/translation/#the-set-language-redirect-view), but we can totally just use a link for this.
> 
> Any good reason not to? then we don't have to do that `onclick` thing.

I think it centered around the form submit actually changing/setting the language in the backend and creating the cookie.  I can look into it again.

pcraig3:
> I think it centered around the form submit actually changing/setting the language in the backend and creating the cookie. I can look into it again.

Yeah, I was wondering if the post request was doing anything in particular, not totally sure. I had the sense that as soon as you hit an "/fr/" domain path, that would set the cookie anyway. If that's true, we should be good to only have a link.

With the syntax error in the "onclick", the form wasn't ever being submitted but the language seemed to be being set properly, so I think we're probably okay to remove the form.

bryan-robitaille:
Changed done:
- Removed the form and only the link remains.  (Tested and language switching works fine - no cookie set)
- Consolidated logic and ensure language is capitalized

profiles/templates/includes/nav_header.html
```diff
@@ -1,21 +1,15 @@
...
-        {% if LANGUAGE_CODE == 'en' %}
-          <input type="hidden" name="language" value="{{ langFr.code }}" />
-          <button>{{ langFr.name_local|title }}</button>
-        {% else %}
-          <input type="hidden" name="language" value="{{ langEn.code }}" />
-          <button>{{ langEn.name_local|title }}</button>
-        {% endif %}
-      </form>
+      <h2 class="visually-hidden">{% trans "Language selection" }</h2>
+      <a role="button" href="{% get_href LANGUAGE_CODE %}">{% get_lang_title LANGUAGE_CODE langEn langFr %}</a>
```
pcraig3, line 11: We can remove `role="button"`. That's only useful when we want the browser to think the link is a button. In this case the link works just fine.

PR #88, opened 2020-07-13
Feature force login after sucessful invitation

PierrePaul:
People are now being logged in after filling the HealcareInviteForm/View. 
There is 2 things I would like someone input on:

* Should the HealthcareView restricted to people **not** logged in?
*  We don't have any tests that covers the last stretch of an invitation (the InviteForm can be filled in and is redirecting). Should I make some?


profiles/forms.py
```diff
@@ -112,7 +112,6 @@ def clean_province(self):
         return HealthcareProvince.objects.get(abbr=province_abbr)
 
     def clean_email(self):
-        print("clean email", self.cleaned_data)
```
pcraig3, line 115: good catch! This was probably me who left this in by accident.

PierrePaul:
> If you mean the `landing` page, I think it's fine to show that to people who are not logged in – they just probably won't see it.

I'm talking about the link you get inside the email. When you click on the link, you are taken to a page to complete registration.
This page is available right now if you are already logged in as well. So, you are completing registration while being logged in.

This means, when you're done completing registration, you will end up logged in as another user.

pcraig3:
> When you click on the link, you are taken to a page to complete registration.

Interesting. This is useful for us in testing and it would be annoying to remove it. I'll make a note of this but let's leave it alone for now.

PR #89, opened 2020-07-13
Style Invitation + 2fa emails

PierrePaul:
Add some styling for invitation and 2fa emails.
Remove reference to svg files and use static images.

Also updated the text with v2 from figma.

Important note, copy writing mention that emails should start with 
`Hello {{ full_name }},`
We are not asking the person name when sending an invitation, so I only kept `Hello,`

Screenshot_2020-07-13 COVID Alert - give numbers to patients - info pierre-paul com - FIXRS Mail

Images are not showing up in local, since gmail can't call localhost:8000


pcraig3:
The images are huge when you see them in a browser. Let's add some CSS to get them to look normal.

Screen Shot 2020-07-14 at 8 45 01 AM

PierrePaul:
> The images are huge when you see them in a browser. Let's add some CSS to get them to look normal.

I totally missed that. Thanks! I also updated to use include, footer.html and header.html for emails.

PR #90, opened 2020-07-13
Security headers

bryan-robitaille:
This PR addresses some of the issues raised on https://observatory.mozilla.org/analyze/covid-alert-portal.alpha.canada.ca.
- Adds content security policy to deny all external resources except those defined in settings.py.
```
CSP_DEFAULT_SRC = ["'self'"]
CSP_STYLE_SRC = ["'self'", "fonts.googleapis.com"]
CSP_FONT_SRC = ["'self'", "fonts.gstatic.com"]
CSP_SCRIPT_SRC = ["'self'", "cdnjs.cloudflare.com"]
```
- Blocks client side JS from accessing CSRF cookie
- Adds HTTP Strict Transport Security for a minimum duration of 6 months.  Basically the browser will not allow to connect to the site with anything but HTTPS for that period of time.  Only active when `is_prod` is true.
- Refactored inline JS so not to throw error from Content Security Policy.  Move to it's own js file and imported value needed using `{{ json_script }}` tag.


bryan-robitaille:
Getting a CSRF token error now when debug fill outs the 2FA form.  When debug is not active everything works perfectly.  Looking into it.

PR #92, opened 2020-07-14
[hotfix] Add missing i18n tag to the password reset page template

pcraig3:
Looks like we missed this earlier.

Causes a template error right now when you try to load the page.


PierrePaul:
:+1:

PR #102, opened 2020-07-15
Feature users UUID

PierrePaul:
Change the user_id to be a random UUID instead of a number auto incrementing.
This needs a drop of the database, since many tables are using User as foreignkey and would need updating (like django-otp).


PierrePaul:
Nice thanks, I will also add unit testing for those so it shows up.

PR #103, opened 2020-07-15
Content Security Policy New Relic fix

bryan-robitaille:
Remove the need to have New Relic inject an inline script at the beginning of the <head> tag.
- Script moved to separate file
- New Relic agent no longer injects inline script with config in `newrelic.ini`
- Removed the `unsafe-inline` option from CSP exemptions in `settings.py`


bryan-robitaille:
Don't merge yet... working on feature to ensure the right browser script is injected for prod and staging

bryan-robitaille:
@pcraig3 let me know if you agree with the changes, then I'll merge if there's no changes requested

profiles/templatetags/new_relic.py
```diff
@@ -0,0 +1,20 @@
...
+@register.simple_tag
+def new_relic():
+    app_name = os.getenv("NEW_RELIC_APP_NAME", "")
+    new_relic_script = ""
+    if app_name not in ["HC_Portal_Staging", "HC_Portal_Production"]:
+        return ""
+    elif app_name == "HC_Portal_Staging":
+        new_relic_script = "js/new_relic_staging.js"
+    else:
+        new_relic_script = "js/new_relic_production"
```
pcraig3, line 18: i think you missed a `.js` here

bryan-robitaille: Getting long in the day.. thanks for fresh eyes.

PR #105, opened 2020-07-15
Feature dashes urls

PierrePaul:
Those two are the only urls I have found to have an underscore in them, for front facing urls.


profiles/urls.py
```diff
@@ -44,7 +44,7 @@
     path("login-2fa/", views.Login2FAView.as_view(), name="login-2fa",),
     path("resend-2fa/", views.Resend2FAView.as_view(), name="resend-2fa",),
     path(
-        "password_reset/",
+        "password-reset/",
         PasswordResetView.as_view(form_class=forms.HealthcarePasswordResetForm),
```
pcraig3, line 48: I think the default auth urls, there's also the `password_reset/done` one: https://github.com/django/django/blob/master/django/contrib/auth/urls.py#L17

PR #110, opened 2020-07-16
Add ENV vars to README + update to `DJANGO_ALLOWED_HOSTS`

pcraig3:
This PR adds all the ENV vars that we use to the README.

It also updates the `ALLOWED_HOSTS` env var to `DJANGO_ALLOWED_HOSTS`. This follows the pattern of some of our other DJANGO settings, I've put the namespace in so that we can see them all beside each other in ENV listings on some of our other platforms (ie, Heroku and AWS).


pcraig3:
Good point, I'll comment them out.

PR #111, opened 2020-07-16
Remove all DB env vars except DATABASE_URL

pcraig3:
We have too many env vars and this makes our code simpler to use.

Existing env vars can be aggregated into the DATABASE_URL string by following this format: https://github.com/jacobian/dj-database-url#url-schema

Updated the docker-compose file as well — pretty sure no one else is using the manual DB vars.


portal/settings.py
```diff
@@ -114,9 +114,8 @@
 # https://docs.djangoproject.com/en/3.0/ref/settings/#databases
 
 default_db_path = os.path.join(BASE_DIR, "db.sqlite3")
-if os.getenv("DATABASE_URL") == "":
```
bryan-robitaille, line 117: I had originally included this because without it the `manage.py migrate` command fails.
The error it throws refers to an improper Database Engine configuration.
```
settings.DATABASES is improperly configured. Please supply the ENGINE value. Check settings documentation for more details.
```

pcraig3: Ah, gotcha. Will add it back then.

pcraig3: I only tried it on startup + reads.

PR #113, opened 2020-07-16
Add DJANGO_ADMINS env var for emails to be mailed about prod errors

pcraig3:
Allows us to set multiple emails by using a comma separated string.

Source:
- https://docs.djangoproject.com/en/3.0/ref/settings/#admins


portal/settings.py
```diff
@@ -52,6 +52,13 @@
 if os.getenv("DJANGO_ALLOWED_HOSTS"):
     ALLOWED_HOSTS.extend(os.getenv("DJANGO_ALLOWED_HOSTS").split(","))
 
+# Admins who get emailed about production errors
+ADMINS = []
+
+if os.getenv("DJANGO_ADMINS"):
+    ADMINS.extend([email.strip() for email in os.getenv("DJANGO_ADMINS").split(",")])
```
bryan-robitaille, line 59: `ADMINS` is looking for a tuple comprised of a name and email `("Foo", "foo@example.ca")` according to the doc you referenced.

PR #121, opened 2020-07-21
SMS 2fa

PierrePaul:
Replace the email 2fa by a SMS 2fa, using Notify as transport.
This goes in pair with https://github.com/PierrePaul/django-otp-notify .
Right now the Pipfile.lock uses version 0.3.0.

There is a new field in the signup process, for the phone number.
Screenshot_2020-07-21 Create account – COVID Alert Portal
With errors
Screenshot_2020-07-21 Error Create account – COVID Alert Portal

When the fields doesn't match
Screenshot_2020-07-21 Error Create account – COVID Alert Portal(3)

I have disabled automatic login after sign up, to force 2fa. We will need to add the flash messages here, to at least let the user know he can login now : 
Screenshot_2020-07-21 Log in – COVID Alert Portal

I updated the templates to remove any mention of Email
Screenshot_2020-07-21 Validate your code — COVID Alert Portal

Note : in the signup form, I am still asking for the HCW email. The email is still used to login, but no longer used for 2fa.


PierrePaul:
The issue is now fixed on heroku. I had to connect manually to DB to figure out the issue. When adding the phone_number field, I was expecting it to set the default to +1647.... but looks like it ended up using `''` as default.
I created a new migration, that would set the real default for users with `''` as phone_number and also create a default sms device.
It now works on heroku, let's cross fingers for staging. I had to manually connect to the DB to figure out the issue and I hope I havent changed something manually.

portal/settings.py
```diff
@@ -206,6 +207,13 @@
 OTP_LOGIN_URL = "login-2fa"
 LOGIN_REDIRECT_URL = "start"
 LOGOUT_REDIRECT_URL = "landing"
+OTP_NOTIFY_ENDPOINT = os.getenv(
+    "OTP_NOTIFY_ENDPOINT", "https://api.notification.alpha.canada.ca"
+)
```
pcraig3, line 212: If the env var exists but is an empty string, then this wont return the second param. That's why we've been using the `or` syntax in other places.

Pipfile
```diff
@@ -28,6 +28,9 @@ newrelic = "*"
 django-axes = "*"
 uwsgi = "*"
 django-csp = "*"
+phonenumbers = "*"
+django-phonenumber-field = "*"
```
pcraig3, line 32: Do we need both of these? Not sure where they're both being used.

portal/settings.py
```diff
@@ -206,6 +207,13 @@
...
+OTP_NOTIFY_API_KEY = os.getenv("OTP_NOTIFY_API_KEY")
+OTP_NOTIFY_TEMPLATE_ID = os.getenv("OTP_NOTIFY_TEMPLATE_ID")
+OTP_NOTIFY_TOKEN_VALIDITY = 30
```
pcraig3, line 215: Can we make this 90 for usability?

portal/settings.py
```diff
@@ -206,6 +207,13 @@
...
+OTP_NOTIFY_NO_DELIVERY = True
```
pcraig3, line 216: Is this ever overridden in other environments?

profiles/forms.py
```diff
@@ -102,9 +104,30 @@ class SignupForm(HealthcareBaseForm, UserCreationForm):
 
     name = forms.CharField(label=_("Full name"), validators=[MaxLengthValidator(200)])
 
+    phone_number = PhoneNumberField(
+        help_text=_(
+            "A single use code will be sent to this phone number everytime you try to login."
```
pcraig3, line 109: We like to say `log in` as two words.

profiles/forms.py
```diff
@@ -102,9 +104,30 @@ class SignupForm(HealthcareBaseForm, UserCreationForm):
...
+        ),
+        widget=forms.TextInput(attrs={"placeholder": "+1"}),
```
pcraig3, line 111: Let's skip adding a placeholder: https://adamsilver.io/articles/placeholders-are-problematic/

A better pattern is to give an example with the hint text (although you're doing it as a validation message), maybe that's fine for now.

profiles/forms.py
```diff
@@ -122,10 +145,24 @@ def clean_email(self):
...
+    def clean_phone_number_confirmation(self):
+        phone_number = self.cleaned_data.get("phone_number")
+        phone_number_confirmation = self.cleaned_data.get("phone_number_confirmation")
+        if (
+            phone_number
+            and phone_number_confirmation
+            and phone_number != phone_number_confirmation
+        ):
+            raise forms.ValidationError(
+                _("The phone number fields doesn't match."), code="invalid",
```
pcraig3, line 157: `don’t match`

pcraig3: Should the code be more specific? `invalid_phonenumber`?

profiles/signals.py
```diff
@@ -0,0 +1,40 @@
...
+def _save_previous_email(sender, instance, **kwargs):
+    try:
+        instance._pre_save_instance = HealthcareUser.objects.get(pk=instance.id)
+    except HealthcareUser.DoesNotExist:
+        instance._pre_save_instance = instance
+
+
+@receiver(post_save, sender=HealthcareUser)
+def update_2fa(sender, instance, created, **kwargs):
+    if not created:
```
pcraig3, line 16: We're not currently allowing the phone number to be updated, eh?

profiles/tests.py
```diff
@@ -32,6 +32,7 @@ def get_credentials(
         "province": province or get_province(),
         "is_admin": is_admin,
         "password": password,
+        "phone_number": "+12125552368",
```
pcraig3, line 35: Is this a fake number?

Pipfile
PierrePaul: Yes, we are adding a phonenumber field to the HealtcareUser. The `django-phonenumber-field` can actually use 2 backends to generate and clean up the value of a phonenumber, so you can choose a more battle-tested/complete solution (phonenumbers) or a more lightweight solution (phonenumberslite).

portal/settings.py
PierrePaul: Of course

profiles/forms.py
PierrePaul: It would, yes! But then it would no longer works with the PhoneNumberField widget. https://github.com/stefanfoulis/django-phonenumber-field/blob/master/phonenumber_field/formfields.py#L21

profiles/signals.py
PierrePaul: not at the moment no :( 
I had to test it via a python shell.

profiles/tests.py
PierrePaul: Kind of. This is the number used as an example by django-phonenumbers-field.
https://github.com/stefanfoulis/django-phonenumber-field/blob/master/phonenumber_field/formfields.py#L31
This is the number from the original Ghostbusters movie.

portal/settings.py
PierrePaul: It's not, it was left there by mistake. Thanks!

PierrePaul:
>     * there's no validation messages showing up on the 2FA form. Are they in the template but they're not coming through?

We have javascript validation in place 

Screenshot_2020-07-22 Valider votre code — Portail Alerte COVID

If you're talking about those errors:
- code expired
- invalid code

They have never been implemented. I can work on that though.

profiles/forms.py
```diff
@@ -102,9 +104,29 @@ class SignupForm(HealthcareBaseForm, UserCreationForm):
...
     name = forms.CharField(label=_("Full name"), validators=[MaxLengthValidator(200)])
 
+    phone_number = PhoneNumberField(
+        help_text=_(
+            "A single use code will be sent to this phone number every time you try to log in."
+        ),
+    )
+    phone_number_confirmation = PhoneNumberField(
+        help_text=_("Enter the same phone number as before, for verification"),
+        widget=forms.TextInput(attrs={"placeholder": "+1"}),
```
pcraig3, line 114: Let's remove the placeholder here too.

profiles/forms.py
```diff
@@ -122,10 +144,24 @@ def clean_email(self):
...
+    def clean_phone_number_confirmation(self):
+        phone_number = self.cleaned_data.get("phone_number")
+        phone_number_confirmation = self.cleaned_data.get("phone_number_confirmation")
+        if (
+            phone_number
+            and phone_number_confirmation
+            and phone_number != phone_number_confirmation
+        ):
+            raise forms.ValidationError(
+                _("The phone number fields don't match."), code="invalid",
```
pcraig3, line 156: Let's use the curly quote: `‘` instead of `'`.

profiles/models.py
```diff
@@ -17,7 +18,13 @@ class Meta:
...
 class HealthcareUserManager(BaseUserManager):
     def create_user(
-        self, email, name, province, is_admin=False, is_superuser=False, password=None
+        self,
+        email,
+        password=None,
+        is_admin=False,
+        is_superuser=False,
+        **kwargs
+        # self, email, password = None, name, province, is_admin = False, is_superuser = False, password = None, phone_number = None
```
pcraig3, line 27: We can remove this comment right?

portal/settings.py
```diff
@@ -206,6 +207,12 @@
 OTP_LOGIN_URL = "login-2fa"
 LOGIN_REDIRECT_URL = "start"
 LOGOUT_REDIRECT_URL = "landing"
+OTP_NOTIFY_ENDPOINT = (
+    os.getenv("OTP_NOTIFY_ENDPOINT") or "https://api.notification.alpha.canada.ca"
+)
+OTP_NOTIFY_API_KEY = os.getenv("OTP_NOTIFY_API_KEY")
+OTP_NOTIFY_TEMPLATE_ID = os.getenv("OTP_NOTIFY_TEMPLATE_ID")
+OTP_NOTIFY_TOKEN_VALIDITY = 90
```
pcraig3, line 216: Since you removed `OTP_NOTIFY_NO_DELIVERY`, the PR review app breaks because the key isn't there (it is in the 'testing' environment). 

We can either add it to the PR environment (pipeline > settings) or remove it when `!is_prod`, so it would only work on staging and prod.

profiles/forms.py
```diff
@@ -102,9 +104,29 @@ class SignupForm(HealthcareBaseForm, UserCreationForm):
 
     name = forms.CharField(label=_("Full name"), validators=[MaxLengthValidator(200)])
 
+    phone_number = PhoneNumberField(
+        help_text=_(
+            "A single use code will be sent to this phone number every time you try to log in."
```
pcraig3, line 109: Okay, I see what the placeholder was for. 

Let's add this. `Include a +1 before your number.`

It will probably change in future but this is MVP -- we should tell people that the code won't validate unless they include the country code.

portal/settings.py
PierrePaul: I ended up using `DEBUG` instead. That's what we are using the form (if DEBUG, put the code directly in the login form).

profiles/views.py
```diff
@@ -102,13 +103,26 @@ def get_initial(self):
...
                 self.request.session[DEVICE_ID_SESSION_KEY] = device.persistent_id
 
-        is_valid = super().form_valid(form)
-        return is_valid
+        # Using the first device, since we don't care which it is, let's
+        # check if the user is being throttled
+        if devices[0]:
+            verified_allowed, _ = devices[0].verify_is_allowed()
+            if verified_allowed is False:
+                form.add_error("code", "Please try again later.")
```
pcraig3, line 116: Trying on the PR review app (entering a different code than what was expected), I'm getting this error instead of the "didn't match" error. 

When do we want this error to be triggered vs the other one?

PierrePaul: I moved the error logic around for the throttling error.
It was always showing this error first, since I was making the check right after an invalid OTC.

PR #123, opened 2020-07-22
Hotfix key generation display

PierrePaul:
Fix the format for live server calls


portal/settings.py
```diff
@@ -206,6 +206,8 @@
 OTP_LOGIN_URL = "login-2fa"
 LOGIN_REDIRECT_URL = "start"
 LOGOUT_REDIRECT_URL = "landing"
+API_AUTHORIZATION = os.getenv("API_AUTHORIZATION")
+API_ENDPOINT = os.getenv("API_ENDPOINT")
```
pcraig3, line 210: This is a great idea. All our `os.getenv` checks should be in settings.

pcraig3:
> Upon receipt of the One Time Code (OTC) in the healthcare portal, there are no data integrity checks to ensure the data is of a proper format. Although the data is trusted as it comes from the key server its still good practice to have some integrity checks to ensure no corruption has taken place. 

> For example in the code below the OTC is split into two 4-byte components. If the received code is less than 8 bytes long and out of bounds access exception will be triggered.

> Update: turns out array slices in python will just extract till the end of the array if the slice specified is larger than its length. Therefore an exception won’t be triggered. Some sanity checks should still take place to ensure the OTC is of the proper format.

profiles/views.py
```diff
@@ -227,10 +227,13 @@ def code(request):
...
-    # Split up the code with a space in the middle so it looks like this: 123 456 789
-    diagnosis_code = (
-        f"{diagnosis_code[0:3]} {diagnosis_code[3:6]} {diagnosis_code[6:10]}"
-    )
+    # Make sure the code has a length of 10, cheap sanity check
+    if len(diagnosis_code.strip()) == 10:
+        # Split up the code with a space in the middle so it looks like this: 123 456 789
+        diagnosis_code = (
+            f"{diagnosis_code[0:3]} {diagnosis_code[3:6]} {diagnosis_code[6:10]}"
+        )
```
pcraig3, line 235: This doesn't do anything otherwise, it just spits it onto the screen. If we did this check on line 222, we could raise some kind of error, log it, and then fall back to ten zeros.

portal/settings.py
```diff
@@ -206,6 +207,16 @@
...
+
+CORS_ALLOW_CREDENTIALS = False
+CORS_ORIGIN_WHITELIST = []
```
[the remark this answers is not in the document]
pcraig3: Nvm, not making Ajax requests, so we should be good to not allow cross-origin requests.

PR #124, opened 2020-07-23
Feature 2fa admin

PierrePaul:
Protects the admin via 2FA SMS.
If the user is not already logged in + verified (via 2fa), the admin ask the user to login via the website.

Screenshot_2020-07-22 Connexion Site d’administration de Django

Something you might find a bit weird in this PR, I have setup the templates DIRS to load portal/templates.
Admins url are set in the main app, `portal` and I needed to override the login template.
I could have moved the admins url to the profiles app, but I feel like this app is already bloated enough as it is.

Creating a new `custom_admin` app just for those modifications didnt feel justified to me, but let me know if you have a different opinion.


portal/templates/admin/login-disabled.html
```diff
@@ -0,0 +1,24 @@
...
+
+{% block nav-global %}{% endblock %}
+
+{% block content_title %}{% endblock %}
+
+{% block breadcrumbs %}{% endblock %}
+
+{% block content %}
+    <div id="content-main">
+        <p>Please use the <a href="{% url 'login' %}">main website to login</a>.</p>
```
pcraig3, line 22: We should put this in `trans` tags, and feel free to add a translation yourself. It's not going to be a high traffic page but it should be in both languages.

portal/urls.py
```diff
@@ -3,6 +3,13 @@
 from django.http import HttpResponse
 from django.conf.urls.i18n import i18n_patterns
 
+from django.utils.translation import gettext_lazy as _lz
+
+from .admin import Admin2FASite
+
+admin.site.__class__ = Admin2FASite
+admin.site.site_header = _lz("COVID Health portal")
```
pcraig3, line 11: Let's call it "COVID Alert Portal administration"

pcraig3: Can we also get the name of the title tag to change?

It would be good if they both said "COVID Alert Portal administration".

image

PR #125, opened 2020-07-23
Add pwned-passwords-django to the app

PierrePaul:
This PR adds pwned-passwords-django module to the app.
Whenever a user tries to sign up with a common password, they will be shown this error : 

Screenshot_2020-07-22 Erreur Créer un compte — Portail Alerte COVID

https://trello.com/c/4KP9Whyp/187-ban-bad-passwords

By default, if HIBP is offline or times out (after 1 second), it will revert to using the django builtin validator CommonPassword, which contains the top 20k password from HIBP, lowercase and deduplicated : https://gist.github.com/roycewilliams/281ce539915a947a23db17137d91aeb7

This is shipped by default with django, but not enabled by default : https://github.com/django/django/blob/master/django/contrib/auth/common-passwords.txt.gz

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -28,6 +28,7 @@ newrelic = "*"
 django-axes = "*"
 uwsgi = "*"
 django-csp = "*"
+pwned-passwords-django = "*"
 
 [pipenv]
 allow_prereleases = true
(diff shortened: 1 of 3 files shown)


pcraig3:
Sorry for not putting my name on the card. I think #126 addresses this.

pcraig3:
I'm going to close this for now. We can revisit later if needed.

PR #131, opened 2020-07-23
Update production branch with latest changes from staging

pcraig3:
Lots in here:

- Added audit trails
- Added 2FA using SMS instead of using email addresses
- Changed the format of the one-time key
- Added a few pages where admin users can se sent invitations
- Updated some environment variable names
  - eg, `ALLOWED_HOSTS` is now `DJANGO_ALLOWED_HOSTS`
- Added an "instructions" page with screenshots

And more as well but those are the big ones.


pcraig3:
💣

PR #132, opened 2020-07-23
Remove /invitations/send-invite URL

pcraig3:
As pointed out in the Blackberry security assessment, regular users can invite other users by manually typing in the URL because we added all of the URLs in the invitations plugin.

> Application uses external library, called django-invitations, to serve invitation feature for administrator and superuser roles. The application includes all views from external library. Two views from django-invitations require only logged-in user to sendinvitations.

This PR only includes the accept invite URL and removes the "create-invite" URL.

## Screenshot

This is the screen we were able to see: 

Screen Shot 2020-07-23 at 6 42 49 PM


PierrePaul:
:+1:

PR #133, opened 2020-07-24
Feature throttling keys

PierrePaul:
Introduce a basic throttling mechanism for the key generation.
I created a new app called covid_key and I moved all key related templates to the new app.
I also moved a few "generic" templates into portal/templates, to try to cleanup the `profiles` app.
Default limit of key per 24h.

I wanted to use easyaudit to check the throttling of the key, but I figured that since I had to make a model for the key anyway for it to show up in easyaudit, I might as well save the user who created the key and not rely on easyaudit. So easyaudit works with the new COVIDKey model, but for calculating the throttling, I just check the keys created by current user in the last 24 hours.

Currently there is no way to reset the throttling on a user, aside from waiting.
I'm also logging via the `logger.error` when a user has hit the threshold.

Screenshot_2020-07-24 Daily limit reached — COVID Alert Portal


PierrePaul:
https://trello.com/c/w3vZJtUt/195-log-number-of-one-time-keys-generated-by-each-account-user

I also added a column in the user admin that shows exactly how many keys were generated by a user in the last 24 hours.

Screenshot_2020-07-24 Sélectionnez l’objet healthcare user à changer COVID Health Portal administration Administration du P

For the full list of keys (even before 24 hours), we can use easyaudit for that, with the filters on right.

Screenshot_2020-07-24 Sélectionnez l’objet CRUD event à changer COVID Health Portal administration Administration du Portai

portal/settings.py
```diff
@@ -249,6 +250,7 @@
 INVITATIONS_INVITATION_EXPIRY = 1  # 1 day
 INVITATIONS_ADMIN_ADD_FORM = "profiles.forms.HealthcareInvitationAdminAddForm"
 INVITATIONS_ADMIN_CHANGE_FORM = "profiles.forms.HealthcareInvitationAdminChangeForm"
+COVID_KEY_MAX_PER_USER_PER_DAY = 1
```
pcraig3, line 253: Lol we should definitely not ship this as the default.

If debug is on, we should default to something really high. Since `0000000000` still counts as a key, making fake requests that don't return actual keys to will still count against your daily total, which could lock you out unexpectedly while developing.

covid_key/apps.py
```diff
@@ -0,0 +1,5 @@
+from django.apps import AppConfig
+
+
+class CovidKeyConfig(AppConfig):
+    name = "covid_key"
```
pcraig3, line 5: Yeah, this is awesome.

covid_key/migrations/0001_initial.py
```diff
@@ -0,0 +1,28 @@
...
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='COVIDKey',
+            fields=[
+                ('id', models.UUIDField(default=uuid.uuid4, editable=False, primary_key=True, serialize=False)),
+                ('key', models.CharField(max_length=12)),
```
pcraig3, line 22: Hmmmmmmm,

So here is the [craic](https://www.urbandictionary.com/define.php?term=What%27s%20the%20craic%3F) with this. 

A concern with our logs is creating systems that associate keys with healthcare workers, because if you can then associate healthcare workers with the people they talk to, you could end up making accurate guesses about which person got which key. Keys are only good for 24 hours and then they get deleted from the server DB, so we don't want to store plaintext keys forever, because then we are keeping them for longer than needed.

SO,

Here's something @maxneuvians suggested earlier in the week.

If we hash the key at the time of creation with the AUTHORIZATION token we're using to connect to the server, then, in the event of a breach, we can send the hashed keys to the server team and then they can hash all the keys they know about to identify duds. 

That means that we never save the literal value of the key to the DB, but always an obscured value. If we get hacked or something, nobody can figure out what keys were shown. 

Does that make sense?

covid_key/templates/covid_key/throttled.html
```diff
@@ -0,0 +1,8 @@
+{% extends 'base.html' %}
+{% load i18n %}
+
+{% block title %}{% trans "Daily limit reached" %}{% endblock %}
+
+{% block content %}
+    <p>{{ _("You have hit your daily limit of code generation.") }}</p>
+{% endblock %}
```
pcraig3, line 8: Easy  👨‍🍳 👌

covid_key/urls.py
```diff
@@ -0,0 +1,9 @@
+from django.urls import path
+
+from . import views
+
+urlpatterns = [
+    # The otp_required is used directly in the view in this case
```
pcraig3, line 6: Let's remove the comment, since it doesn't make sense in this context.

covid_key/views.py
```diff
@@ -0,0 +1,79 @@
...
+        f"{diagnosis_code[0:3]} {diagnosis_code[3:6]} {diagnosis_code[6:10]}"
+    )
+
+    expiry = timezone.now() + timedelta(days=1)
+
+    covid_key = COVIDKey()
+    covid_key.created_by = request.user
+    covid_key.expiry = expiry
+    covid_key.key = diagnosis_code
+    covid_key.save()
```
pcraig3, line 67: Let's create this object before we do the `f"{diagnosis_code[0:3]} {diagnosis_code[3:6]} {diagnosis_code[6:10]}"` transformation.

If we end up hashing the values, we should preserve the original value we got back from the API, not a differently-formatted version.

portal/settings.py
pcraig3: `100 if is_prod else 1000000`?

portal/settings.py
```diff
@@ -270,8 +272,8 @@
     AXES_ENABLED = False
     AXES_VERBOSE = False
     AXES_LOGGER = "axes.watch_login"
-    logger = getLogger(AXES_LOGGER)
-    logger.setLevel(CRITICAL)
+    getLogger(AXES_LOGGER).setLevel(CRITICAL)
+    getLogger("covid_key.views").setLevel(CRITICAL)
```
pcraig3, line 276: Nice, this is great.

profiles/views.py
```diff
@@ -1,9 +1,5 @@
-import requests
-import logging
-from datetime import timedelta
-
 from django.conf import settings
-from django.shortcuts import render, redirect
+from django.shortcuts import redirect
 from django.contrib.sites.shortcuts import get_current_site
 from django.utils.translation import gettext as _
```
pcraig3, line 4: So nice to clean all this up. 👍

pcraig3:
Overall this is a 10/10 PR though, love the refactoring! 🏅

covid_key/migrations/0001_initial.py
maxneuvians: I think this make sense. I would recommend using bcrypt to hash the IDs. I would however also run this past @obrien-j because of the classification of the OTK.

PierrePaul: It makes a lot of sense, thanks to both of you

PierrePaul: I updated the PR. I used scrypt, since it has been included in the python standard library since python 3.6, https://docs.python.org/3/library/hashlib.html#hashlib.scrypt it requires openssl 1.1+.

The call looks like this : `scrypt(password=diagnosis_code.encode(), salt=token.encode(), n=16384, r=8, p=1)` 

Im using the autorization key as salt, if no key is set (like in local), it uses the SECRET_KEY as the salt.

Let me know if you have a better idea. `bcrypt` is available, but it's not included in the standard library.

PierrePaul: Note, Im saving the key as binary right now, but if it makes more sense, I could save only the hexdigest.

pcraig3: Talked to @obrien-j and he recommended we not save the key at all. If we see some suspicious activity and we want to cancel keys, we can provide a list of timestamps to the server team and they can almost for sure be able to identify the right keys.

covid_key/migrations/0001_initial.py
```diff
@@ -0,0 +1,28 @@
...
+import uuid
+
+
+class Migration(migrations.Migration):
+
+    initial = True
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+    ]
```
pcraig3, line 15: There's a note on the Django "migrations" documentation about using the historical models. (ie, we want the user model at this point in history, not the most recent user model.)

https://docs.djangoproject.com/en/3.0/topics/migrations/#historical-models

Not sure what `swappable_dependency` is, but does it account for this?

pcraig3: Sounds like we're good here.

PR #134, opened 2020-07-26
AWS ready docker build image / Terraform config / FARGATE deployment files

bryan-robitaille:
This PR includes all the terraform config to reproduce the Production environment.  Security specs are identical to the covid-server except for the egress from Fargate on ports 443 and 587.  This egress is required for the app to submit statistics to New Relic and contact Notify for SMS (443) as well as 587 for SMTP.  Logs are all streamed to Cloud Watch from containers and include separate log groups for app containers and VPC traffic.

GitHub workflows included a Terraform security scan and docker build and image push to AWS ECR.
Removed automatic deployment to Production environment.  Rolling back to manual for security reasons.

Let's setup a time to go through this in the coming week.  I'm on vacation but will find the time.


portal/templates/base.html
```diff
@@ -1,12 +1,17 @@
...
-{% load i18n %}
+{% load static i18n new_relic settings %}
 {% get_current_language as LANGUAGE_CODE %}
-{% load new_relic %}
 <html lang="{{ LANGUAGE_CODE }}">
   <head>
     {% new_relic %}
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
+
+    {% with sha="GITHUB_SHA"|get_setting %}
```
PierrePaul, line 9: cool :)

PR #135, opened 2020-07-27
Remove autocomplete attributes from inputs

pcraig3:
The BB security review flagged the autocomplete attribute as a risk.

The autocomplete attribute prompts the browser to remember the login details of the account that logged into the portal.

Since we don't want accounts on shared computers to be remembered, this is no good for us.

Sources:
- https://stackoverflow.com/questions/2580955/disable-autocomplete-on-textfield-in-django


PierrePaul:
:+1:

PR #137, opened 2020-07-29
Feature contact api

PierrePaul:
https://trello.com/c/mwL7FLJy/78-people-can-submit-feedback-via-a-form-on-the-portal

Adds a contact form that sends information to Freshdesk.
Screenshot_2020-07-29 Contact us — COVID Alert Portal


contact/forms.py
```diff
@@ -0,0 +1,48 @@
...
+    )
+
+    def send_freshdesk(self):
+        try:
+            response = post(
+                settings.FRESHDESK_API_ENDPOINT + "/tickets",
+                json={
+                    "email": self.cleaned_data.get("email", None),
+                    "subject": self.cleaned_data.get("feedback", "")[0:20],
+                    "description": self.cleaned_data.get("feedback", ""),
```
pcraig3, line 33: Keep the first 20 characters for the subject line, right?

contact/forms.py
```diff
@@ -0,0 +1,48 @@
...
+                    "name": self.cleaned_data.get("name", ""),
+                    "product_id": settings.FRESHDESK_PRODUCT_ID,
+                    "status": 2,
+                    "priority": 1,
+                },
+                auth=(settings.FRESHDESK_API_KEY, "dummyString"),
```
pcraig3, line 39: Is there a reason to keep "dummyString" rather than just having it return `""` or `None`?

contact/forms.py
```diff
@@ -0,0 +1,48 @@
...
+            )
+            response.raise_for_status()
+        except Exception as e:
+            logger.critical(f"Unable to communicate with Freshdesk : {e}")
+            raise ValidationError(
+                _(
+                    "We are unable to send your feedback at the moment. Please try again."
+                )
+            )
```
pcraig3, line 48: Interesting. I'll ask @sidewalkballet if there's a better message we can give here.

contact/templates/contact/form.html
```diff
@@ -0,0 +1,15 @@
...
+    {% trans "Contact us" %}
+{% endblock %}
+
+
+{% block content %}
+    {% include  "includes/error_list.html" %}
+
+    <h1>{% trans "Contact us" %}</h1>
+    {% include "includes/cds_form.html" with submit_text=_("Send") %}
+{% endblock %}
```
pcraig3, line 15: So nice having these small templates.

contact/forms.py
pcraig3: Let's do this for now. `There was a problem and your feedback wasn't submitted. Please email your feedback to assistance+covid@cds-snc.ca.`

contact/tests.py
```diff
@@ -0,0 +1,34 @@
...
+
+        post_data = {
+            "name": "dummy name",
+            "email": "test@test.com",
+            "feedback": "[test please delete] feedback",
+        }
+
+        response = self.client.post(reverse("contact:index"), post_data,)
+        self.assertEqual(response.status_code, 302)
+        self.assertEqual(response.url, reverse("contact:success"))
```
pcraig3, line 34: Could this send a real request, eg if you're running it locally but you have the right env vars?

pcraig3: nvm, saw down below that it wouldn't.

contact/views.py
```diff
@@ -0,0 +1,24 @@
...
+    template_name = "contact/form.html"
+    form_class = ContactForm
+    success_url = reverse_lazy("contact:success")
+
+    def form_valid(self, form):
+        try:
+            # It's a bit sad, but if we send to freshdesk when running tests (which happens a lot),
+            # Freshdesk will spam everyone email about "A new ticket has been created"
+            if not settings.TESTING:
+                form.send_freshdesk()
```
pcraig3, line 19: Okay, so during testing you would see the success message but nothing is sent?

pcraig3:
I also have a few comments about the form itself. 

1. Let's remove the colons. I've done this in `/profiles/forms.py` with a BaseHealthcareForm, so maybe import it?

2. The textarea could use some work.

- we still have [rounded edges when the textarea is focused](image
), so let's remove them (it's either `webkit-appearance` or `border-radius`)
- if there's an error, let's do a red border, like this: https://design-system.service.gov.uk/components/textarea/#error-messages
- if there's an error AND it's focused, then do a thicker black border and a yellow outline.

textarea

3. Let's do the same thing with the inputs if they have an error. Right now we have yellow AND red AND black, so let's just do a thick black border. Examples all over the GDS design system: https://design-system.service.gov.uk/components/text-input/

contact/forms.py
PierrePaul: Yeah the subject is a mandatory field for the freshdesk api, so Im taking the first 20chars. I can add "..." at the end though

contact/forms.py
PierrePaul: The API docs says I need to send something as password, but I havent tested sending nothing.

contact/views.py
PierrePaul: Exactly :(

contact/forms.py
PierrePaul: Yup it works with empty string. Resolving.

pcraig3:
One more thing I forgot: I know it's not in the mocks for some reason, but let's put `Contact us` as a link in the main nav, after "Start" and "Manage accounts"

locale/fr/LC_MESSAGES/django.po
```diff
@@ -18,8 +18,68 @@ msgstr ""
...
-#: covid_key/templates/covid_key/key.html:9
+#: contact/forms.py:12
+#| msgid "Full name"
+msgid "Your name"
+msgstr "Nom complet"
+
+#: contact/forms.py:14
+#| msgid "Email address"
+msgid "Your email address"
+msgstr "Adresse courriel"
```
pcraig3, line 29: In the figma file, these are different, but I don't think it's a substantial difference. happy to leave these how you have them.

portal/templates/includes/nav_main.html
```diff
@@ -4,6 +4,7 @@
   <div class="page--container">
     <ul>
       <li><a {% if request.resolver_match.url_name == 'start' %}class="active"{% endif %} href="{% url 'start' %}">{% trans "Start" %}</a></li>
+      <li><a {% if request.resolver_match.url_name == 'start' %}class="active"{% endif %} href="{% url 'contact:index' %}">{% trans "Contact us" %}</a></li>
```
pcraig3, line 7: Let's put this after "Manage accounts". Also the "match" name should match "contact"

PR #138, opened 2020-07-29
Fix the invitation form in admin. Provide a default subject

PierrePaul:
Provide a default for the subject line, when creating an Invitation from the admin.
Also fixes creating invitation from the admin.

https://trello.com/c/rftgddue/197-bug-inviting-a-user-through-the-admin-portal-sends-an-email-with-no-subject-heading


profiles/templates/invitations/email/email_invite_subject.txt
```diff
@@ -1 +1,3 @@
-{{ subject_line }}
+{% load i18n %}
+{% trans "Your portal account" as subject %}
+{{ subject_line|default:subject }}
```
pcraig3, line 3: Can we send the Site name in the subject line like we do with the public-facing one?

Having the name of the environment in the subject line makes the invite clearer in your email inbox.

image

profiles/forms.py
```diff
@@ -207,7 +207,8 @@ def save(self, *args, **kwargs):
         }
         instance = Invitation.create(**params)
         instance.send_invitation(self.request)
-        super().save(*args, **kwargs)
+        # We can't call InvitationAdminForm here, it would try to send 2 invitations
+        super(forms.ModelForm, self).save(*args, **kwargs)
```
pcraig3, line 211: Yeah, good call.

profiles/templates/invitations/email/email_invite_subject.txt
PierrePaul: Yeah I can make a templatetag that would return the site name in there. Thanks

PR #143, opened 2020-07-30
Feature one hour timeout

PierrePaul:
https://trello.com/c/7trzZMGf/153-sessions-should-log-out-a-user-after-1-hour-of-inactivity

Logs you out after 1 hour of inactivity.


portal/templates/base.html
```diff
@@ -1,12 +1,11 @@
 <!DOCTYPE html>
-{% load static %}
-{% load i18n %}
+{% load static i18n new_relic settings %}
```
pcraig3, line 2: Oh neat, I didn't know about this syntax.

portal/templates/base.html
```diff
@@ -1,12 +1,11 @@
...
 {% get_current_language as LANGUAGE_CODE %}
-{% load new_relic %}
 <html lang="{{ LANGUAGE_CODE }}">
   <head>
     {% new_relic %}
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="Refresh" content="{{ "SESSION_COOKIE_AGE"|get_setting|add:1 }}; url={% url 'session_timed_out' %}">
```
pcraig3, line 8: Let's check for a user first before adding this. If there's no request user, you could reset someone who's on the privacy page or contact page, for example.

pcraig3: Although, I gotta say that this is a pretty neat way of doing it.

portal/templates/includes/flash_messages.html
```diff
@@ -0,0 +1,11 @@
+{% load i18n %}
+
+{% if messages %}
+    <div class="messages--container" role="alert">
+        <ul class="messages">
+        {% for message in messages %}
+            <li{% if message.tags %} class="{{ message.tags }}"{% endif %}>{{ message }}</li>
+        {% endfor %}
+        </ul>
+    </div>
```
pcraig3, line 10: I know it's kind of silly, but is there any way to only show this if it's logout message?

Otherwise you see other flash messages about things you did. (we should fix that eventually but not a high priority). Eg, here are some messages you get when inviting users.

Screen Shot 2020-07-31 at 3 33 20 PM

Alternatively, we just remove those other messages.

profiles/static/scss/components/forms/_messages.scss
```diff
@@ -0,0 +1,15 @@
+main .messages {
+  &--container {
+    max-width: 450px;
+    padding: $space-sm;
+    background: rgba(78, 136, 185, 0.1);
+    border: 1px solid #4E88B9;
```
pcraig3, line 6: Small point, but the text is dark blue in the mock.  Pretty sure we can use the same colours we're using in the "tag" for this: https://github.com/cds-snc/covid-alert-portal/blob/main/profiles/static/scss/components/_tag.scss#L13

image

profiles/views.py
```diff
@@ -205,3 +205,12 @@ class UserDeleteView(Is2FAMixin, ProvinceAdminDeleteMixin, DeleteView):
...
     success_url = reverse_lazy("profiles")
+
+
+def redirect_after_timed_out(request):
+    messages.add_message(
+        request,
+        messages.INFO,
+        _("Your session timed out. Log in again to continue using the portal."),
+    )
+    return redirect(reverse_lazy("login"))
```
pcraig3, line 216: This is a great solution to this.

PierrePaul:
PR updated. Im only showing the logout messages now. I also added a unit test about that.

PR #149, opened 2020-08-04
Feature update content 0.4

PierrePaul:
Update the content to iteration 0.4.
Also fix the docker setup when running locally.

Aside from small changes (removing the product nav bar when not logged in), only the content should be affected.

https://trello.com/c/I1ck0jqO/217-update-content-to-v04-and-add-translations-to-the-portal
https://trello.com/c/qyNApEtx/223-remove-the-landing-page-from-the-application
https://trello.com/c/adN7QyM1/209-change-content-in-header

This PR also moves the /start page into the covid_key app.
It also removes the /landing page.


pcraig3:
Added a commit to put more spacing under the login form. Put a bit of explanation in the commit message, but basically we want to consistently separate form items, the "forgot password" link is conceptually related to the form, but the sentence after is not, so push that one further away.

| before | after |
|--------|-------|
| Screen Shot 2020-08-04 at 5 16 14 PM   |   Screen Shot 2020-08-04 at 5 16 11 PM |

covid_key/urls.py
```diff
@@ -1,7 +1,17 @@
...
 
 urlpatterns = [
     path("key/", views.code, name="key"),
+    path(
+        "start/",
+        login_required(
+            otp_required(TemplateView.as_view(template_name="covid_key/start.html"))
+        ),
+        name="start",
+    ),
```
pcraig3, line 16: good move, this makes sense.

portal/templates/includes/cds_form.html
```diff
@@ -24,7 +24,7 @@
     {% endfor %}
   </div>
 
-  <button type="submit">
+  <button type="submit" class="{{ submit_class }}">
```
pcraig3, line 27: Is this getting used? I don't see it anywhere.

PierrePaul: not in v.04. It might be used in v.05, will remove for now. They are planning to make some button red, when it's a destructive operation, but they aren't sure.

PierrePaul:
> Since the card explicitly says "V.04", we can merge without it and I can make a new card, or else you can update that page. Let me know which you went for.

I updated the invite complete page to v0.5.

PR #150, opened 2020-08-04
Don't create a CovidKey object if there was a failure

PierrePaul:
Moving the covidkey logic around to prevent creating an obj when there is an error

https://trello.com/c/mlZ4q1X1/230-dont-create-a-key-object-unless-the-key-retrieval-worked


PierrePaul:
Screenshot_2020-08-04 Give patient this key — COVID Alert Portal(1)

covid_key/views.py
```diff
@@ -31,31 +33,54 @@ def code(request):
 def _generate_key(request):
     token = settings.API_AUTHORIZATION
     diagnosis_code = "0000000000"
+    covid_key = None
     if token:
+        print(token)
```
pcraig3, line 38: Let's remove this.

covid_key/views.py
```diff
@@ -31,31 +33,54 @@ def code(request):
 def _generate_key(request):
     token = settings.API_AUTHORIZATION
     diagnosis_code = "0000000000"
```
pcraig3, line 35: We can also default this to `""`, since we don't want to display the zeroes.

PierrePaul: In development Im still showing the 000 000 0000.
As soon as token and api_endpoint is provided, the error will be shown instead.

pcraig3: gotcha, sounds good.

PR #156, opened 2020-08-06
Require a POST request to see the "key" page

pcraig3:
This PR is makes generating keys more intentional. Since we ask for a new code every time someone loads the `/key` page, you could send someone who is already logged in the URL and they would generate a new codes perhaps unwittingly. 

Or, as @PierrePaul pointed out:

> One of the attack vectors could be a rogue browser extension that knows the path to the key generation. It would be able to call the page until the user has reached the throttle limit and then the user would be stuck, without being able to generate new keys.

This is something Mark brought up, that a CSRF attack (ie, repeatedly loading the page in an authenticated browser), someone could hit their limit of generated codes for the day.

By making "Generate key" into a form rather than a link, we can make sure that generating codes is more intentional. 
- If you try to enter the `/key` url directly in your browser you will be redirected back to the /start page
- Borked POST requests will fail now with a CSRF error
- You can't bookmark the "generate a key" page

The specific changes included are: 
- Adds a form instead of a link to the "start" page
- Redirects from "/key" if not a POST request
- Adds a default 403 CSRF template 
- Fixes the problem where the logged-in header would still say "Log in"

Mostly you don't notice anything has changed.

What you should test is:
- Trying to enter the `/key` url directly in your browser (you should be redirected)
- Messing with the CSRF token on the `/start` page and then submitting the form

Notes:

- On the error message template, the request context wasn't coming through and the header nav was breaking (the language switching link wants the request path). Didn't know how to fix it, so I removed the header nav.
- Added some okay content and french translations to the CSRF page. Don't really want it to go into the Figma as it's very edge-case-y so let's just merge if it looks okay.
- Not sure if the last test (`test_CSRF_page_for_bad_posts`) adds any value. Happy to remove it.

## Screenshots

CSRF 403 Error template

| en | fr |
|--------|-------|
|  Screen Shot 2020-08-06 at 2 21 32 PM  |   Screen Shot 2020-08-06 at 2 22 06 PM  |


PierrePaul:
:+1:

locale/fr/LC_MESSAGES/django.po
```diff
@@ -186,19 +186,33 @@ msgid ""
...
 
+#: portal/templates/403_csrf.html:4 portal/templates/403_csrf.html:7
+msgid "Forbidden"
+msgstr "Interdit"
+
+#: portal/templates/403_csrf.html:9
+msgid "Something went wrong when visiting this page, likely a CSRF violation."
+msgstr ""
+"Un problème est survenu lors de la visite de cette page, probablement une "
+"violation du CSRF."
```
sboisvert, line 197: Allegedly CSRF is "injections de requêtes illégitimes par rebond"
@PierrePaul est ce que c'est utiliser comme terminologie?

locale/fr/LC_MESSAGES/django.po
```diff
@@ -186,19 +186,33 @@ msgid ""
 "try again after 5 minutes."
 msgstr ""
 
+#: portal/templates/403_csrf.html:4 portal/templates/403_csrf.html:7
+msgid "Forbidden"
+msgstr "Interdit"
+
+#: portal/templates/403_csrf.html:9
+msgid "Something went wrong when visiting this page, likely a CSRF violation."
```
sboisvert, line 194: We should probably tell the user what to do. Like, "Please try again from the start page with a link to the start page"

locale/fr/LC_MESSAGES/django.po
PierrePaul: Non et pour avoir généré une erreur CSRF, il faut généralement l'avoir cherché. Donc c'est plus fréquent de voir le terme CSRF plutôt que la traduction complète.

locale/fr/LC_MESSAGES/django.po
PierrePaul: > I don't see a way someone could get that key even if they could get an authenticated user to generate a new one.
> Either way, just explaining the "Why" in the PR would be nice :)

One of the attack vector could be a rogue browser extension that knows the path to the key generation. It would be able to call the page until the user has reached the throttle limit and then the user would be stuck, without being able to generate new keys.

pcraig3: > Why are we doing this change?
> Either way, just explaining the "Why" in the PR would be nice :)

Good point! Updated the description with the problem we are trying to solve.

pcraig3:
Checked it with @sidewalkballet, gonna merge.

PR #157, opened 2020-08-06
Re-add buildspec for Elastic Beanstalk + Add service IP to allowed hosts for ALB healthcheck

pcraig3:
This PR fixes 2 things:

- Our Elastic Beanstalk automated deployments
- Our Fargate manual deployments

### Elastic Beanstalk automated deployments

We updated the `buildspec.yml` file in #134 so that the app will build in ECS, but it broke our builds in our staging environment. 

Since we're still using staging for now, I've re-added the previous EB-compatible buildspec file and updated both EB deploy pipelines to look for this other buildspec file instead of the new Fargate buildspec.

## Fargate manual deployments

Also updated the healthcheck URL for the load balancers in our terraform configuration. The landing page has been removed, so healthchecks to the previous URL will return 404s, not 301s.

I've also added the  service IP to ALLOWED_HOSTS for ALB healthchecks on startup. The target group healthchecks currently fail because of Django's host header validation. ELB/Target groups use the IP of the target as the host header so lets grab it on startup if we're running in a prod environment.

Pretty sure this makes #158 redundant.

### DIFF
--- a/buildspec-eb.yml
+++ b/buildspec-eb.yml
@@ -1,0 +1,33 @@
+version: 0.2
+
+env:
+  variables:
+    DJANGO_ENV: "development"
+phases:
+  install:
+    runtime-versions:
+      python: 3.x
+  pre_build:
+    commands:
+      - pip install --upgrade pip setuptools pipenv
+      - pipenv install --dev
+      - echo "Compile CSS"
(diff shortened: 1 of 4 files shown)


pcraig3:
Fixed the unused import ☝️

PR #158, opened 2020-08-06
add service IP to ALLOWED_HOSTS for ALB healthchecks

stonith:
The target group healthchecks currently fail is [Django does host header validation](https://docs.djangoproject.com/en/3.0/topics/security/#host-header-validation). ELB/Target groups use the IP of the target as the host header and we should be able to derive the instance IP from the AWS metadata service.

### DIFF
--- a/portal/settings.py
+++ b/portal/settings.py
@@ -13,6 +13,7 @@ https://docs.djangoproject.com/en/3.0/ref/settings/
 import os
 import sys
 import ast
+import requests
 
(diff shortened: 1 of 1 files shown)


pcraig3:
Hi @stonith, check this one out: https://github.com/cds-snc/covid-alert-portal/pull/157/commits/d89b6782fd1d9209c6c6efb83199b91198e68a17

I think this PR isn't needed anymore.

PR #163, opened 2020-08-11
Provices a phonetic alphabet for the code generation

PierrePaul:
https://trello.com/c/UhCmwoKT/163-provide-phonetic-alphabet-instructions-for-the-one-time-key
This adds a nice phonetic alphabet that works both in french and english.

Screenshot_2020-08-11 Donnez la clé suivante au patient — Portail Alerte COVID


covid_key/templates/covid_key/key.html
```diff
@@ -11,6 +11,9 @@
       <h1>{% trans "Give patient this key" %}</h1>
       <div id="big-code">
         <code>{{ code }}</code>
+        <hr />
+        {% include "covid_key/phonetic.html" %}
+        <hr />
```
pcraig3, line 16: Tried this out using Voiceover and it says "horizontal splitter" on these. 

Let's do `<hr aria-hidden="true" />` to hide them from the screen readers.

covid_key/templates/covid_key/phonetic.html
```diff
@@ -0,0 +1,14 @@
...
+    <p>{% blocktrans %}To read this phonetically:{% endblocktrans %}</p>
+    <div class="phonetic-wrapper">
+        {% for letter in code %}
+            {% with letter|phonetic as word %}
+                <span>{{ word }}</span>
+            {% endwith %}
+            {% if letter == ' ' %}
+                <br />
+            {% endif %}
+        {% endfor %}
```
pcraig3, line 12: The markup you get here is a bit weird because it inserts extra spans. 
ie, you'll get something like this.

```html
<span>Three</span>
<span>Four</span>
<span>Five</span>
<span></span>
<br />
```

I don't know why, but with the extra span, voiceover was re-reading each line twice.  "Three. Four. Five." "Three. Four. Five."

I rewrote it like this instead:

```
<div class="phonetic-wrapper">
        <div>
        {% for letter in code %}
            {% if letter == ' ' %}
            </div><div>
            {% else %}
                {% with letter|phonetic as word %}
                    <span>{{ word }}</span>
                {% endwith %}
            {% endif %}
        {% endfor %}
        </div>
    </div>
```

And it works much better in my testing (Chrome and Safari), plus doesn't include the extra span.

covid_key/templatetags/alphabet.py
```diff
@@ -0,0 +1,46 @@
...
+    sentences = {
+        "0": _("Zero."),
+        "1": _("One."),
+        "2": _("Two."),
+        "3": _("Three."),
+        "4": _("Four."),
+        "5": _("Five."),
+        "6": _("Six."),
+        "7": _("Seven."),
+        "8": _("Eighth."),
```
pcraig3, line 18: `Eight`! No second "h". 

You would say "eighth" when there it's an order (eg, "the eighth person to finish"), otherwise you say "eight" (eg, "there are eight contestants")

locale/fr/LC_MESSAGES/django.po
```diff
@@ -177,6 +178,106 @@ msgid "You have hit your daily limit of code generation."
...
+msgid "Six."
+msgstr "Six."
+
+#: covid_key/templatetags/alphabet.py:17
+msgid "Seven."
+msgstr "Sept."
+
+#: covid_key/templatetags/alphabet.py:18
+msgid "Eighth."
+msgstr "Huit."
```
pcraig3, line 215: This is still the wrong translation, but we can get to it later if not in this PR.

PierrePaul: Updated, sorry.

PR #164, opened 2020-08-11
All IPs in django-easy-audit request logs say they are from 127.0.0.1.

PierrePaul:
X-Forwarded-For should be an IP, not a domain. Using the HTTP_X_REAL_IP instead of REAL_IP directly

https://trello.com/c/khRD4u2A/270-bug-all-ips-in-django-easy-audit-request-logs-say-they-are-from-127001


covid_key/views.py
```diff
@@ -18,6 +18,7 @@
 @login_required
 @otp_required
 def code(request):
+    print(request.META)
     if request.method != "POST":
```
pcraig3, line 22: No "print"

PR #165, opened 2020-08-11
[bugfixes] Redirect to /start if logged in + stop sending duplicate 2FA codes on account creation

pcraig3:
This PR:

- Redirects to "/start" from the root domain if the user is logged in (previously, it would always load the login page)
- Removes the duplicate SMS TFA code that would be sent on account creation

### Redirect to start page on root if already logged in

If logged in, redirect to start page. Else, redirect to login page.

### Remove explicit 2FA call after account created

Creating a new account triggers `update` in signals.py, which sends a 2FA SMS to the new phone number.
We were also explicitly creating an SMS to send at the end of user creation.
The result was that users would get 2 TFA codes on account creation, but only the second one was usable.

Removing the account-creation call from the view, leaving the signals.py one.

### DIFF
--- a/profiles/tests.py
+++ b/profiles/tests.py
@@ -112,6 +112,21 @@ class AdminUserTestCase(TestCase):
         session.save()
 
 
+class RootView(AdminUserTestCase):
+    def setUp(self):
+        super().setUp(is_admin=True)
+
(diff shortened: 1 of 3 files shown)


pcraig3:
Fixed them ☝️

PR #166, opened 2020-08-12
Add /status response + ignore it in django-easy-audit

pcraig3:
This PR

- Adds a super basic status page
- Tells terraform to use the status page as the health check page
- Tells django-easy-audit to not listen to requests to the status page
- Adds a couple of simple tests


PierrePaul:
:+1:

PR #167, opened 2020-08-12
Feature restrict invitations to 25 per 59 minutes*

PierrePaul:
https://trello.com/c/JJQGYsfL/266-restrict-provincial-admins-to-25-invites-an-hour

This PR adds a generic Throttled Mixin. You can specify any Model to lookup for, any time span and if the user has hit the limit specified, it will trigger `self.limit_reached` method of your class or the default one.

This new generic Throttled Mixin is being use the covid_key page and in the invitation form page.

This PR also fixes the CSRF on the key generation page. We were sending a CSRF, but not validating it.


portal/settings.py
```diff
@@ -270,6 +270,8 @@
 INVITATIONS_ADMIN_ADD_FORM = "profiles.forms.HealthcareInvitationAdminAddForm"
 INVITATIONS_ADMIN_CHANGE_FORM = "profiles.forms.HealthcareInvitationAdminChangeForm"
 COVID_KEY_MAX_PER_USER_PER_DAY = 100 if is_prod else 10000
+MAX_INVITATIONS_PER_PERIOD = 5
+MAX_INVITATIONS_PERDIOD_SECONDS = 300  # 5 minutes
```
pcraig3, line 274: Let's do 25 invites per hour.

profiles/templates/invitations/templates/locked.html
```diff
@@ -0,0 +1,8 @@
+{% extends 'base.html' %}
+{% load i18n %}
+
+{% block title %}{% trans "Invitation limit reached" %}{% endblock %}
+
+{% block content %}
+    <p>{{ _("You have hit your limit of invitation. Please try again in a few minutes.") }}</p>
```
pcraig3, line 7: `You have hit your limit of invitations. Please try again in an hour.`
There might be more content from Yedida to replace this with at some point.

portal/settings.py
pcraig3: `MAX_INVITATIONS_PERDIOD_SECONDS` also has a typo.

Interesting that this one is in settings but the COVID KEY time limit is directly in the view. It feels like they should all be in settings.

profiles/views.py
```diff
@@ -175,6 +179,9 @@ def form_valid(self, form):
         self.request.session["invite_email"] = invite.email
         return super().form_valid(form)
 
+    def limit_reached(self):
+        return render(self.request, "invitations/templates/locked.html", status=403)
+
```
pcraig3, line 184: This is really clean 👍

profiles/templates/invitations/templates/locked.html
pcraig3, line 7: Small nitpick, but all templates should have a header where the H1 matches the title, so let's also do `<h1>"Invitation limit reached"</h1>`

portal/templates/throttled/locked.html
```diff
@@ -0,0 +1,8 @@
+{% extends 'base.html' %}
+{% load i18n %}
+
+{% block title %}{% trans "Daily limit reached" %}{% endblock %}
+
+{% block content %}
+    <p>{{ _("You have hit your limit.") }}</p>
+{% endblock %}
```
pcraig3, line 8: Let's also add an `h1 here as well.

PierrePaul: I have added a H1 to it, but Im still debating if I should keep that template.
I created that template as a default template in case a class is using the mixin and is not providing his own template. In our case, right now, it's not useful. Should I just remove it until we need it?

profiles/templates/invitations/templates/locked.html
PierrePaul: I updated the text with the one provided from Yedida last night.

portal/templates/throttled/locked.html
pcraig3: I think leave it in.

pcraig3:
Merging!

PR #180, opened 2020-08-17
Adds a new api_key field to province. 

PierrePaul:
This new field is used to make the call to COVID alert server to generate the one time key. If the province has no api_key, it will uses the default one provided by the system

https://trello.com/c/oyDDKKvN/276-add-a-token-field-to-the-provinces-table

Screenshot_2020-08-17 Modification de healthcare province COVID Health Portal administration Administration du Portail Aler


profiles/models.py
```diff
@@ -2,12 +2,14 @@
...
 from django.utils.translation import gettext as _
 from django.contrib.auth.models import BaseUserManager, AbstractBaseUser
+from django.conf import settings
 from phonenumber_field.modelfields import PhoneNumberField
 
 
 class HealthcareProvince(models.Model):
     name = models.CharField(max_length=100, unique=True)
     abbr = models.SlugField(max_length=3, allow_unicode=True, unique=True)
+    api_key = models.CharField(null=True, blank=True, max_length=256)
```
pcraig3, line 12: Can we call it a "Bearer token" in the admin interface? (If it's not trivial to change the label, then let's leave it as is).

profiles/admin.py
```diff
@@ -12,13 +12,14 @@
 
 class ProvinceAdmin(admin.ModelAdmin):
     list_display = ["id", "abbr", "name"]
```
pcraig3, line 14: Is it possible to show the token column but with a "true/false" type value in the list view of the admin "list" interface? That way we could see at a glance which provinces we have set tokens for.

If not then let's double-check with Courtney if it's okay to show them on that page. I think it's probably better to.

profiles/models.py
PierrePaul: It is trivial, will update.

profiles/admin.py
PierrePaul: It is trivial to do, will update! Thanks!

PR #183, opened 2020-08-18
Bug language switching

PierrePaul:
Inspired from the setlang function provided by i18n, but without forcing a POST request to change language.
Also inspired from Bryan dual domain templatetags to support different domains based on language.
Threw support for ?next parameter and some unit testing as well.
https://trello.com/c/dgF7MFr2/286-bug-switching-languages-is-imperfect


portal/settings.py
```diff
@@ -314,6 +314,7 @@
 # Site Setup for Separate Domains
 
 URL_DUAL_DOMAINS = os.getenv("URL_DUAL_DOMAINS", "False") == "True"
+# URL_DUAL_DOMAINS = True
```
pcraig3, line 317: Can be deleted

profiles/views.py
```diff
@@ -264,3 +269,52 @@ def redirect_after_timed_out(request):
...
+    # Take the referer by default
+    next_url = urlparse(request.GET.get("next", request.META.get("HTTP_REFERER")))
+    # but if a ?next_url has been provided, let's make sure it's clean
+
+    next_url = next_url.path and unquote(next_url.path)
+    if not url_has_allowed_host_and_scheme(
+        url=next_url,
+        allowed_hosts={request.get_host()},
+        require_https=request.is_secure(),
+    ):
```
pcraig3, line 293: Does this ever happen? When do we get a "next" that's also a language switching URL?

profiles/views.py
```diff
@@ -264,3 +269,52 @@ def redirect_after_timed_out(request):
...
+            settings.URL_FR_PRODUCTION if lang == "fr" else settings.URL_EN_PRODUCTION
+        )
+
+    next_url = root_domain + translate_url(next_url, lang)
+    response = HttpResponseRedirect(next_url)
+
+    if hasattr(request, "session"):
+        request.session[LANGUAGE_SESSION_KEY] = lang
+
+    print(f"setting cookie with {lang}")
```
pcraig3, line 309: 👍  (let's remove it)

profiles/views.py
```diff
@@ -264,3 +269,52 @@ def redirect_after_timed_out(request):
...
+    response.set_cookie(
+        settings.LANGUAGE_COOKIE_NAME,
+        lang,
+        max_age=settings.LANGUAGE_COOKIE_AGE,
+        path=settings.LANGUAGE_COOKIE_PATH,
+        domain=settings.LANGUAGE_COOKIE_DOMAIN,
+        secure=settings.LANGUAGE_COOKIE_SECURE,
+        httponly=settings.LANGUAGE_COOKIE_HTTPONLY,
+        samesite=settings.LANGUAGE_COOKIE_SAMESITE,
```
pcraig3, line 318: We don't have these settings. Do they default to `None` or something else?

profiles/views.py
PierrePaul: I dont think we need it. I just ported it from i18n setlang.

profiles/views.py
PierrePaul: they are set in django but can be overridden : https://docs.djangoproject.com/en/3.0/ref/settings/#language-cookie-httponly

profiles/views.py
pcraig3: I would prefer to if we don't need it because then our code is cleaner, but not too bothered either way.

Feel free to change or merge as-is. Gonna approve.

PR #185, opened 2020-08-19
Metrics alarms

bryan-robitaille:
This PR contains terraform code for the creation of metrics and associated alarms for:
| Metric | Alarm |
| :-----: | :-----: |
| Application Error | Raised after any app error |
| OTK Generation | Raised after 11 in 1h |
| Site table change | Raised after any change |
| Account Lockout | Raised after 5 lockout events in 1h |
| Invitation Lockout | Raised after any lockout event |

This PR also reduces the threshold for blocking requests by IP from 250 / 5 minutes to 100 / 5 minutes for Requests to /login and POST method.

When alarms are raised they are delivered through a Lambda function to the exposure healthcare dev Slack channel.  This is only applicable to the staging environment.


config/terraform/aws/cloudwatch.tf
```diff
@@ -106,6 +106,110 @@ resource "aws_cloudwatch_metric_alarm" "application_error_warn" {
 }
 
 
+###
+# AWS CloudWatch Metrics - Activity Alarms
+###
+
+resource "aws_cloudwatch_log_metric_filter" "key_generation" {
+  name = "KeyGeneration"
+  pattern = "\"CRUD event_type:CREATE model:covid_key.covidkey\""
```
obrien-j, line 115: Is there a clever way to do this per api-key or are we limited to system wide metrics?

bryan-robitaille: Unfortunately CloudWatch metrics can't aggreate based on log entries.  A separate metric would need to be created for each province/territory and the log output modified to include which province/territory generated the token.

PR #186, opened 2020-08-21
Dont send sms on login. 

PierrePaul:
Dont send SMS on login if the user has a Yubikey set.
Redirect to Yubikey verify on password reset if user has a Yubikey

### DIFF
--- a/profiles/views.py
+++ b/profiles/views.py
@@ -59,7 +59,7 @@ class YubikeyVerifyView(FormView):
 
     def get_form_kwargs(self):
         kwargs = super().get_form_kwargs()
-        device = RemoteYubikeyDevice.objects.filter(user=self.request.user).first()
+        device = self.request.user.remoteyubikeydevice_set.first()
         # Pass the device to the form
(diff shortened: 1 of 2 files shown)


profiles/views.py
```diff
@@ -174,8 +174,8 @@ def get(self, request, *args, **kwargs):
         if request.user.is_verified():
             return redirect(reverse_lazy("start"))
 
-        if request.user.is_superuser:
-            yubikey = RemoteYubikeyDevice.objects.filter(user=request.user).first()
+        if request.user.remoteyubikeydevice_set.first() is not None:
+            yubikey = request.user.remoteyubikeydevice_set.first()
```
pcraig3, line 178: We're doing the same check here twice: `request.user.remoteyubikeydevice_set.first()`, in the `if` condition and then right afterwards.

Initially I did the `is_superuser` check so that we wouldn't do a DB call if the user wasn't a superuser but it's probably cleaner to just use the one `if` and then redirect if we find it.

PR #187, opened 2020-08-21
Delete invitation of an user when deleting the user

PierrePaul:
Some fields are not editable even by managers, so managers delete a user then invite them again.
This PR delete all invitation tied to the email of the user being deleted, so Managers can invite them again.

https://trello.com/c/gd7qVgST/296-admins-should-be-able-to-send-invites-to-previously-deleted-accounts


pcraig3:
Looks good, tried it on Heroku.

PR #188, opened 2020-08-24
Set remote IP header for easy audit

bryan-robitaille:
Sets the header so that Easy Audit lists the actual remote IP and not that of the Load Balancer.


pcraig3:
Gonna merge + deploy this to staging so that we know it works 👍

PR #192, opened 2020-08-25
Changing the default value of OTP_THROTTLE_FACTOR to 3

PierrePaul:
https://trello.com/c/OogOgaQF/306-update-otp-throttle-factor
Changes the default value of `1` to `3` for the throttling factor (the waiting time between each failed attempts).


pcraig3:
Gonna merge while it's updated.

PR #194, opened 2020-08-25
Fix the password urls and forms

PierrePaul:
Overwrites the url name to make sure the proper form is loaded and the proper url is used.

https://trello.com/c/pI9pKUN8/304-make-sure-that-the-forms-inherit-properly


profiles/urls.py
```diff
@@ -88,6 +88,7 @@
 )
 # Django.contrib.auth urls have underscore in them, let's change that for dashes
 urlpatterns += [
+    path("", include("django.contrib.auth.urls")),
```
pcraig3, line 91: Doesn't going first mean that it has precedence?

PierrePaul: It does, but then we overwrite the route named 'password_reset' and make it point to our form instead.

pcraig3:
The "new password" form having the colons. We want to get rid of the 'autocomplete' attributes that are in there.

image

PierrePaul:
> The "new password" form having the colons. We want to get rid of the 'autocomplete' attributes that are in there, and inheriting from the base form removes that for us.

Since are overwriting all endpoints of django.contrib.auth.urls, I removed the include. I added a new Form that inherits HealthcareBaseForm to remove the `:` in the Confirm Reset Password view.

PR #197, opened 2020-08-27
Add a skip link

pcraig3:
This PR: 

- adds a skip link
- bumps the footer links a little

### adds a skip link

It's an a11y best practice and we should just put it in. If you hit "tab" right away after the page loads, you will see the skip link. If you then hit "enter", it will focus on the `<main>` content.

If you're not logged in, there's no menu to speak of, so we don't need to "skip to the content". 

Doesn't target the "main" element itself, because we want to skip the "logged-in menu" if it exists.

Lifted it directly from what @brdunfield did on [a past PR for CRA](https://github.com/cds-snc/cra-claim-tax-benefits/pull/410).

### bumps the footer links

Ever since we removed the blue dots, the links are slightly too close on big screens. Now they have a bit more space.

## gif

skip4


portal/templates/base.html
```diff
@@ -30,6 +30,9 @@
     <title>{% block title %}{% trans "COVID Alert Portal" %}{% endblock %}{% block title_suffix %} {% trans "— COVID Alert Portal" %}{% endblock %}</title>
   </head>
   <body>
+    {% if request.user.is_authenticated %}
+      <div id="skip-link-container"><a href="#page--content" id="skip-link">Skip to main content</a></div>
```
bryan-robitaille, line 34: Any technical reason why we're not translating 'Skip to main content'?

PR #199, opened 2020-08-27
Feature authorized domains

PierrePaul:
Add a safe list of domains that can be sent invitations to.

https://trello.com/c/hJnWwOFf/283-limit-account-invites-to-certain-email-domains

Screenshot_2020-08-27 Error Add an account — COVID Alert Portal(2)


profiles/forms.py
```diff
@@ -277,6 +278,23 @@ def validate_invitation(self, email):
...
+            domain = email[email.find("@") + 1 :]
+            AuthorizedDomain.objects.get(domain=domain)
+        except AuthorizedDomain.DoesNotExist:
+            if settings.DEBUG is False or settings.TESTING is True:
+                raise forms.ValidationError(
+                    _(
+                        "You cannot invite %(email)s to create an account because @%(domain)s is not on the portal’s safelist. To add @%(domain)s to the safelist, email healthcare-en@cds-snc.ca."
+                    )
+                    % {"domain": domain, "email": email}
+                )
```
pcraig3, line 295: A few things here:

- settings.DEBUG isn't granular enough because it's on in staging and prod, so we probably need a new setting (ie, something we can turn on only in prod). The side-effect of that would be that we can test it locally too. 
- I've messaged Brian about the message, and he's updating it.
- The email we've been giving people on other pages is `assistance+healthcare@cds-snc.ca`

profiles/migrations/0011_auto_20200827_1428.py
```diff
@@ -0,0 +1,79 @@
...
+        'toronto.ca',
+        'wdgpublichealth.ca',
+        'wechu.org',
+        'york.ca',
+    ]
+    for domain in domains:
+        try:
+            AuthorizedDomain.objects.get(domain=domain)
+        except AuthorizedDomain.DoesNotExist:
+            AuthorizedDomain.objects.create(domain=domain)
```
pcraig3, line 56: This is cool. Do we want superadmins to be able to add them ad-hoc, or do we want to only allow them in via migrations? (@brianhendrick1 you're acting PM).

If we allow superadmins to add new ones, we won't have all the emails in the codebase but it will be easy to update.
If we don't allow superadmins to add new ones, we can keep them all in the codebase of the repo, so we can always know who is allowed to be invited.

pcraig3: Also do we want "cds-snc.ca" domains to be allowed? This is for the production site only, not the staging site, so I would tend to think we don't want them allowed.

brianhendrick1: I will get back to you ASAP on the first question. Need to better understand and assess both options.

Regarding second question, no need to allow cds domains. From a security perspective, the less domains allowed the better - since no "cds-snc.ca" domains will be HCW providing one-time keys to patients there is no need to have them in production (besides the superadmins).

PierrePaul: That was also my understanding as well. Also keep in mind that we can create users from the admin. Invitation forms should be used by managers.

PierrePaul: The migration is run only one time. Im making a check here just in case someone decides to rollback the migration.

brianhendrick1:
If it is not possible to have two font weights in the same message, let's remove the second sentence and keep it in the error validation message.

Updated design can be viewed [in Figma](https://www.figma.com/file/znPqLTQQDcUkaNjdDcShiz/Exposure-Notification---Health-Portal-(mock-up)?node-id=2497%3A44).

profiles/migrations/0011_auto_20200827_1428.py
brianhendrick1: Let's leave it with superadminess for now, and we can assess whether we stick with superadmins or migrations when Courtney gets back.

I created a [trello card](https://trello.com/c/FfUqSrw4/313-assess-the-mechanism-to-safelist-email-domains-through-superadmins-or-migrations) to action this decision.

profiles/forms.py
```diff
@@ -277,6 +278,23 @@ def validate_invitation(self, email):
...
+            domain = email[email.find("@") + 1 :]
+            AuthorizedDomain.objects.get(domain=domain)
+        except AuthorizedDomain.DoesNotExist:
+            if settings.DEBUG is False or settings.TESTING is True:
+                raise forms.ValidationError(
+                    _(
+                        "You cannot invite %(email)s to create an account because @%(domain)s is not on the portal’s safelist."
+                    )
+                    % {"domain": domain, "email": email}
+                )
```
pcraig3, line 295: I think we need a new environment variable for this:

- using `settings.DEBUG` will be the same in staging and prod, but in staging we want to be able to invite people arbitrarily
- (minor point, but) it's hard to test this locally because we need DEBUG to be `True` when running on localhost

Is there a better way to do this?

PR #200, opened 2020-08-31
Remove New Relic client side settings

bryan-robitaille:
Given that Chrome blocks these requests now given new browser security settings we might as well remove it.


PierrePaul:
Since we are not really using it server side, should we remove it as well?

bryan-robitaille:
> Since we are not really using it server side, should we remove it as well?

Just to reference what we talked about during out Dev standup and close the loop in this conversation.  Keeping New Relic for the Server Side metrics could be beneficial in the long run when we want to start refactoring after all major features are done.

pcraig3:
I think it's worth a note in the README that we're using server-side New Relic but nothing client-side.

https://github.com/cds-snc/covid-alert-portal/blob/72a6484d95d2854433fc78017425bb74b612d1ba/README.md#new-relic-configuration

PR #206, opened 2020-09-02
Moves the validation error on confirmation fields to the first field

PierrePaul:
Moves the error message of all fields that needs a confirmation (phone_number, password) into the first field.

https://trello.com/c/ZR6ePNoX/196-a11y-password-validation-error-during-signup-should-be-on-the-first-password-field

Screenshot_2020-09-02 Error Change your mobile phone number — COVID Alert Portal

Screenshot_2020-09-02 Error Create account — COVID Alert Portal

Screenshot_2020-09-02 Error Create account — COVID Alert Portal(1)


pcraig3:
A couple comments on this one:

1. It looks like you can reset your password through your account with 2 passwords that don't match. (The first one will be used). Also, you are logged out immediately after changing your password, which is maybe not the intended behaviour.

2. On the "reset password" page, the alert is still on the second field.

Screen Shot 2020-09-03 at 11 06 14 AM

pcraig3 (changes requested):
I think the code is good now (it works!!), but can you push a patch version update as well and then merge.

Feels like a bugfix to me — actually several.
-  making the system work better for some users
- don't lose the session after password changes
- updating passwords now looks for mismatches

PierrePaul:
I updated the changelog and version. https://github.com/cds-snc/covid-alert-portal/pull/206/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR15

I didnt mention the `updating passwords now looks for mismatches`, since it was a bug introduced by the PR.

pcraig3:
Ah, parfait!

pcraig3 (approved):
Approve! Thanks for this update!

PR #207, opened 2020-09-02
Change Slack alert icons depending on severity

bryan-robitaille:
Just cause I wanted to get cute while killing time.
Alerts coming into slack will use 1 of 3 available icons as it's avatar.

If everything if A-Ok and this is just informational: 📢   (aka the loudspeaker)
If this is a warning and something you might want to look into: ⚠️  (aka warning sign)
If this is a critical alert, so drop your coffee and scramble back to the pc: 🚨  (aka rotating light)

These icons are supplementary to the message color bands that will appear based on the same severity criteria (green, orange, and red)


pcraig3 (approved):
📢 Seems fun 📢

PR #208, opened 2020-09-03
Feature block users

PierrePaul:
https://trello.com/c/4hV2nCBu/323-add-ability-for-superadmins-to-disable-all-other-users

Let superuser (admins) block other user via the django-admin interface.
Includes a hotfix for the new LoginHandler. Now it's properly checking if the user exists instead of throwing a 500 error.

The admin change user form also removes the is_admin, is_superuser (and the new blocked) fields if the logged in user is editing his own user, to prevent admins from locking themselves out.


pcraig3 (commented):
Sorry if I'm coming late to this, but couldn't we use [`is_active`](https://docs.djangoproject.com/en/3.1/ref/contrib/auth/#django.contrib.auth.models.User.is_active) instead? 

It's already part of our user model, but we've never used it for anything (so it seems like this would be a good fit).

also, if you are "blocked" by putting in bad emails, it doesn't register in the admin as blocked — which is the intended behaviour, just wanted to write it down.

PierrePaul:
You're right! Removed the migration and using is_active now.

pcraig3:
Okay, so I have a number of follow-ups.

1. I'm getting an error creating a new user now because "blocked" is being set to a non-null value (ie, try to add a user through the admin portal).

2. The "blocked until" section is required, so you can't submit an update to a user without setting it.

Screen Shot 2020-09-03 at 4 35 07 PM

3. The "blocked until" section isn't being respected. Ie, if someone is inactive, but "blocked until" a year ago (ie, 2019-09-03), then they still get in.

I think the easiest thing to do here, (despite my earlier comment 💡 ) is to remove the "blocked_until" field. Users are active or not as a checkbox and there's no notion of how long they are blocked surfaced in the UI.

pcraig3:
Okay, so I talked to Courtney about it and she says all good to keep the datefield. 

Tested it again on Heroku and locally. Saw the error on Heroku but not locally (I guess migrations aren't run again after the PR is also made.

A couple of comments:
- Can we put in a couple of tests to show that logging in as an `is_active=False` account fails or if `blocked_until` exists (and is in the future)?
- When I tried it locally, it said there's a new migration because we changed the `blocked_until` on the model, so I have a new migration file on my local one. Should probably be committed 😇

Overall this is really good though. The logic looks solid 👍

PierrePaul:
Thanks! Never hesitate to ask for more tests! :)

pcraig3 (approved):
Looks good to merge! 👍

PR #211, opened 2020-09-03
Lower the limit of onetime key generation per day to 25

PierrePaul:
https://trello.com/c/mRNV2h6O/336-reduce-keys-per-day-throttling-limit-to-25

Just lowering the 100 key per day limit to 25 in production.


pcraig3 (approved):
👍

PR #212, opened 2020-09-08
Add aria-describedby to form inputs if validation errors 

pcraig3:
### Add aria-describedby attribute if validation errors exist

"aria-describedby" attributes tell screen readers to read more contextual information about a given field.

If someone is using a screenreader and they are pressing "tab" to get to different form fields, they should be made aware when they reach a field with a validation error.

This way, the validation error will be read after the form label.

## gif

error


PierrePaul:
Looking good, should we update the changelog + version?

pcraig3:
Yup, will do.

PierrePaul (approved):

PR #214, opened 2020-09-09
Removing requirements.txt to force snyk to read the Pipfile instead

PierrePaul:
Right now the Snyk github action picks up the file requirements.txt under the `portal` directory, but this file is not used.
We are using the Pipfile to generate the python packages, so removing this file then force Snyk to re-import the project and pick up the Pipfile, when the file is in the master branch.


pcraig3 (approved):
🪓

pcraig3:
Merging this closes #209

PR #215, opened 2020-09-09
Fix snyk

PierrePaul:
Adds a Snyk scan with Github action.


bryan-robitaille (commented):
Looking good.  Ran like it was supposed to and targeted the Pipfile.

bryan-robitaille (approved):

pcraig3:
If you look at the "successful checks", it looks like they're doubling up, but the "(ExposureNotification)" scan only seems to be looking at terraform files, not the project Pipfile.

image

PR #217, opened 2020-09-10
[DO NOT MERGE] Use vulnerable django version

pcraig3:
This is just to test what the Snyk warning looks like (and which job catches it).


pcraig3:
Closing because we got what we needed out of this.

PR #218, opened 2020-09-14
Two-column login page with "Thank you" message

pcraig3:
Add in a second column to our login page with a message for healthcare workers.
It's a nice visual message when the screen real-estate is available, but once we have constraints we should hide it. 

- hide it on mobile: it doesn't help you log in 
- hide it for screen readers (same as above): it doesn't include any information that helps you log in

The images don't have alt tags on purpose because we don't want screenreaders to be aware of the right-hand side of the screen.

## Screenshot

Screen Shot 2020-09-14 at 12 06 49 PM


PierrePaul:
It looks good and feel great, aside from missing french.
The top right language switcher kind of feel weird floating, not aligned to the right, but I understand the rest of the content fit inside that fixed size container.

pcraig3:
@PierrePaul, I agree about the floating French header, but it's at the same breakpoint as other canada.ca pages are, so it's pretty standard. The design makes it a bit awkward though, because not much else outside of the footer actually gets that far across the page.

brianhendrick1 (approved):
Looks great! Nice work team :)

PierrePaul (approved):

PR #219, opened 2020-09-14
Set the created_by to null when deleting the linked user. 

PierrePaul:
Also keeping a log of the original user email.
Incident response.


pcraig3 (changes requested):
This isn't a huge emergency to get out, so I think we should make this a more typical PR.

Let's:

- add a test that we can delete a user who has created a key
- update the CHANGELOG file

Can happen tomorrow if this evening we're done work.

pcraig3 (approved):
It's working for me locally now 👍 

If we have too many more tests like that, we should look at a teardown method that would reset the throttle limit, but this works for now.

PR #220, opened 2020-09-15
Adding check in Lambda to parse JSON + enable logging

bryan-robitaille:
This PR adds the ability for Lambda to create a new Log Group in CloudWatch and Log Steam to that Group.
A 🐛 fix that ensures the Alarm message is a JSON before trying to extract a key.  This is needed because of the 2 types of triggers for the Alarms (Metrics vs Events) that have different structures.


PierrePaul (approved):
LGTM! :partying_face:

PR #223, opened 2020-09-17
New welcome page and updated quick guide

PierrePaul:
Add a new welcome page to help Healthcare Workers get a feeling of the generate code page without pressing the "Generate key" button.

The welcome page is only shown once, after the signup, but before the 2fa.

https://trello.com/c/is3F7lex/321-new-staff-accounts-can-gain-familiarity-with-the-portal-without-generating-real-otks

Screenshot_2020-09-17 Welcome to the portal — COVID Alert Portal

Screenshot_2020-09-17 Welcome to the portal — Portail Alerte COVID

PR #228, opened 2020-09-18
(New!) New welcome page and updated quick guide #223

pcraig3:
All the hard work is done in #223, this just cleans a few things up.

This PR adds the welcome pages (see description of #223 for the screencaps) as well as a couple of fixes:

- Fixes the layout bug where the example tax would overlap the example key
- The welcome page is now shown after 2fa, rather than after the "signup" form 
- Demotes a couple of `h2`s to `h3`s

## gif

Here's an example of the example text collapsing on the welcome page. 

code

Merging this closes #223


PierrePaul (approved):
Looking good to me!

PR #231, opened 2020-09-21
Fix the CSS for the /key page + bump version

pcraig3:
When I updated the CSS for the welcome page and the quick guide, I accidentally broke the CSS for the "key" page. No harm done as the information is still on the page, but it's a bit harder to read through.

Took the opportunity to consolidate some of our templates.

Bumping the version as part of this PR so that we can release it faster.

## Screenshots

| before | after |
|--------|-------|
|  removed the CSS, so the information is all still there but fairly unstyled   |  added back the CSS we had before  |
|  Screen Shot 2020-09-21 at 2 46 01 PM   |  Screen Shot 2020-09-21 at 2 46 04 PM |


bryan-robitaille (approved):
Ready to Roll!

PR #232, opened 2020-09-21
Fix main nav layout on login page

pcraig3:
On the login page, we zeroed out all the max-width spacing so that we could have full-bleed columns, but when you are on the login page _and you are logged in_, the nav menu is touching the side of the screen, which is a total oversight.

By making our CSS rules more specific, we fix the layout glitch.

## Screenshots

| before | after |
|--------|-------|
|  the "main" nav elements break our column layout   |   the "main" nav elements are aligned with the logo and the rest of the content    |
|  Screen Shot 2020-09-21 at 3 34 16 PM    |  Screen Shot 2020-09-21 at 3 34 18 PM   |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 - Title Case the words in the phonetic alphabet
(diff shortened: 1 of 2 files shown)


PierrePaul (approved):
lgtm!

PR #234, opened 2020-09-22
Rename alerts to be more generic

bryan-robitaille:
See file diff for 1 liner changes.


pcraig3 (approved):
I think this looks fine.

PR #235, opened 2020-09-22
Fix the login screen image on IE11

pcraig3:
The thank you hand breaks out of the top of the column without an explicit height when you're on IE11.

Also added a `min-height` to the blue column 

### ie11 login page

| before | after |
|--------|-------|
|  "thank you" message breaks out of the column     |  "thank you" message is centred     |
|  Screen Shot 2020-09-22 at 3 30 14 PM  |  Screen Shot 2020-09-22 at 3 29 59 PM    |

### login page on hi-res screens

| before | after |
|--------|-------|
|  blue column looks dumb on big screens   |  blue column is more normal looking     |
| Screen Shot 2020-09-22 at 4 24 20 PM  | Screen Shot 2020-09-22 at 4 24 30 PM  |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Title Case the words in the phonetic alphabet
 - Fixed CSS for "logged-in" nav items for the login page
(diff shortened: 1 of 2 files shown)


bryan-robitaille (approved):
🎉

PR #236, opened 2020-09-23
Only upload Prod containers when Version file is updated

pcraig3:
This pull request changes our deployment action so that we only upload containers to production if the Version file is updated.

We used to have one action that would: 

- Run a terraform security check
- Build a container in the local workspace
- Sign in to Staging Amazon and push the container to the container registry
- Sign in to Prod Amazon and push the container to the container registry

But I've split it out into 3 actions.

1. Run terraform scan & build container in local workspace
2. Wait for 1, then sign in to Staging Amazon and push the container to the container registry
3. Wait for 1, then check if "Version" file updated. If so, sign in to Prod Amazon and push the container to the container registry

### DIFF
--- a/.github/workflows/prod-build-push-container.yml
+++ b/.github/workflows/prod-build-push-container.yml
@@ -1,0 +1,85 @@
+name: Prod — Push container to ECR
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "VERSION"
+
+env:
+  ECR_REPOSITORY: covid-alert-portal-terraform
+  GITHUB_SHA: ${{ github.sha }}
+
+jobs:
+  push-prod:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Wait for container to pass Django CI 3.6 Test
+        uses: fountainhead/action-wait-for-check@v1.0.0
+        id: wait-for-CI-tests-3_6
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          checkName: test (3.6)
+          ref: ${{ github.sha }}
+
+      - name: Wait for container to pass Django CI 3.7 Test
+        uses: fountainhead/action-wait-for-check@v1.0.0
+        id: wait-for-CI-tests-3_7
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          checkName: test (3.7)
+          ref: ${{ github.sha }}
+
+      - name: Wait for container to pass Django CI 3.8 Test
+        uses: fountainhead/action-wait-for-check@v1.0.0
+        id: wait-for-CI-tests-3_8
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          checkName: test (3.8)
+          ref: ${{ github.sha }}
+
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Read VERSION file
+        id: version
+        uses: juliangruber/read-file-action@v1
+        with:
+          path: ./VERSION
+
+      - name: Build Covid Portal
+        run: docker build -t base --build-arg GITHUB_SHA_ARG=$GITHUB_SHA .
+
+      - name: Configure Production AWS credentials
+        id: aws-production
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.PROD_AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.PROD_AWS_SECRET_ACCESS_KEY }}
+          aws-region: ca-central-1
+
+      - name: Login to Production Amazon ECR
+        id: login-ecr-prod
+        uses: aws-actions/amazon-ecr-login@v1
+
+      - name: Tag Images for Production
+        env:
+          ECR_REGISTRY: ${{ steps.login-ecr-prod.outputs.registry }}
+          VERSION: ${{ steps.version.outputs.content }}
+        run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/staging-deploy.yml
```diff
@@ -6,7 +6,6 @@ on:
 
 jobs:
   deploy-portal-service:
-    if: github.ref == 'refs/heads/main' && github.event_name == 'push'
```
pcraig3, line 9: Pretty sure this is just duplicating the conditional at the top of the file:

```
on:
  push:
    branches: [main]
```

.github/workflows/build-container.yml
```diff
@@ -0,0 +1,57 @@
+name: Build container
+
+on:
+  push:
+    branches: [main]
+
+jobs:
+  terraform-security-scan:
```
bryan-robitaille, line 8: Thinking we can drop this `terraforom-security-scan` as it's already running in the Terraform.yml flow on any PR's that touch the `config/terraform/aws` path as well as any pushes to `main`

.github/workflows/build-container.yml
```diff
@@ -0,0 +1,57 @@
...
+
+      - name: Wait for container to pass Django CI 3.8 Test
+        uses: fountainhead/action-wait-for-check@v1.0.0
+        id: wait-for-CI-tests-3_8
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          checkName: test (3.8)
+          ref: ${{ github.sha }}
+
+      - name: Checkout
```
bryan-robitaille, line 47: Not sure if the remaining actions `Checkout, Read VERSION file, Build COVID Portal` are still required in this file.
The `Build COVID Portal` action is duplicated in both the prod-push-container and staging-build-push-container

.github/workflows/build-container.yml
```diff
@@ -0,0 +1,57 @@
...
+        uses: actions/checkout@v2
+      - name: Terraform security scan
+        uses: triat/terraform-security-scan@v2.0.1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  build-container:
+    needs: terraform-security-scan
+    runs-on: ubuntu-latest
+    steps:
```
bryan-robitaille, line 22: These 3 wait actions should be moved to `prod-push` and `staging-build-push` and this file can be deleted.  The two build and push actions are currently not waiting for the Python checks to pass.

.github/workflows/prod-push-container.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Login to Production Amazon ECR
+        id: login-ecr-prod
+        uses: aws-actions/amazon-ecr-login@v1
+
+      - name: Tag Images for Production
+        env:
+          VERSION: ${{ steps.version.outputs.content }}
+        run: |
+          docker tag base $PROD_ECR_REGISTRY:$GITHUB_SHA
+          docker tag base $PROD_ECR_REGISTRY:$VERSION
```
bryan-robitaille, line 54: Nice!  I didn't think of loading the version file to tag the image.

bryan-robitaille (changes requested):
I think we can tweak these a little more.  Let's chat tomorrow during our Dev stand up to quickly go over some possible changes.

.github/workflows/build-container.yml
pcraig3: Okay, removed.

.github/workflows/build-container.yml
pcraig3: You're right. I was thinking that we could save time by building it once. I think (🤔 ) that the container is built in the same place, so building it twice in the staging + prod actions is wasted effort.

.github/workflows/build-container.yml
pcraig3: Pushed an update to test if waiting starts working.

.github/workflows/prod-push-container.yml
pcraig3: 👍  it's great! been working so far.

.github/workflows/build-container.yml
pcraig3: You were right all along. New update does what you suggested.

.github/workflows/build-container.yml
pcraig3: Nope, you were right. Removed this file.

.github/workflows/prod-build-push-container.yml
```diff
@@ -0,0 +1,83 @@
...
+
+on:
+  push:
+    branches: [version-pr]
+    paths:
+      - "VERSION"
+
+env:
+  GITHUB_SHA: ${{ github.sha }}
+  PROD_ECR_REGISTRY: ${{ secrets.PROD_ECR_REGISTRY }}
```
pcraig3, line 11: @bryan-robitaille, I was wondering why this one is different than the staging one though. Ie, here the value is a secret vs in staging we wait for the output of a previous step. Shouldn't we do it the same way in both files?

bryan-robitaille: Was just following the Server teams precedent in which everything regarding Prod seems to be secret.

pcraig3: Yup, all good. If we get the variable from the past command, we can remove the secret in this repo.

pcraig3:
Okay, all good. Tested this merging onto another branch and both jobs ran. Should be good to merge.

bryan-robitaille (approved):
Awesome!  Let's get this merged 🎉

PR #238, opened 2020-09-24
Add back styles.css and Remove Route53 HealthCheck

bryan-robitaille:
- Add styles.css by modifying `.gitignore`.  This file was mistakenly left out in the original PR and showed how bad a maintenance screen looks like without CSS. 
- Remove unused/unrequired `css` file.
- Remove HealthCheck and modify Route 53 Failover policy to leverage default Load Balancer check instead of custom.  https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-values-failover-alias.html


pcraig3 (approved):
Good if you are. 👍

PR #241, opened 2020-09-25
Align new wait with correct dependency

bryan-robitaille:
Rename await dependency in Deploy to Staging action to new version in Push to Staging.  Variable name was `build-n-push` and is now `push-staging`


pcraig3 (approved):
Thanks, Bryan!

PR #243, opened 2020-09-28
Adding a migration to fix invitation with a sent field set to None

PierrePaul:
Having invitation with a sent field set to None can have unexpected results, like in django-admin or when listing the invitation, depending on the use.
This PR adds a migration that fix all previously created invitations that doesn't have a sent value.


pcraig3 (approved):

pcraig3:
Merci 😇

PR #245, opened 2020-09-28
Delete previous invites if no user account exists

pcraig3:
Previously, we were allowing:

- invite user
- user would sign up (invite is marked "accepted")
- user is deleted
- user is re-invited
- invite fails because accepted invite already exists

We pretty much always want to be able to invite someone when there is no user account for that person, so this PR makes that change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


pcraig3:
ps, it's not possible to test this unless you can manually adjust the database, as there's currently no way to get an "accepted" invite for a non-existent user.

So if you are testing, you should do it locally, which is what I did.

bryan-robitaille (approved):
Looks good to me :)

PR #246, opened 2020-09-29
Standardization of alarm and filter resource names

bryan-robitaille:
Standardizing on alarm names ending in `_warn` or `_critical` as well as Alarm names not containing hyphens.  Filters for metrics do not contain the `_warn` or `_critical` suffix.


pcraig3 (approved):
👍

PR #248, opened 2020-09-29
Update the instructions on the "key" page

pcraig3:
This PR:

- updates the instructions on the "key" page, including changing all the pictures
- updates the FR info/warning banner on the "start" page

To see the changes, click the "generate key" button on this page: https://covid-alert-v6-key-c7c9ezaasm0.herokuapp.com/en/start/

(Note that when you switch between FR and EN on the "key" page itself, it will return you to the start page. This is intentional behaviour.)

## Screenshots

It's hard to see because the screen is so long but here's how the top of it looks.

| before | after |
|--------|-------|
| Screen Shot 2020-09-29 at 4 20 46 PM   | Screen Shot 2020-09-29 at 4 20 42 PM   |


YedidaZalik:
looks good to me, thanks @pcraig3

bryan-robitaille (approved):
Looking good to me.  Checked it out in English and French.

pcraig3:
Removing the version bump as I can get a few more content changes in before releasing.

PR #249, opened 2020-09-29
Adding Routing rules and Cache Control to Failover Solution

bryan-robitaille:
Using Cloudfront custom errors to catch 403 (Forbidden) response from S3 Origin and return `/en.htm` path for all files/urls that don't exist.

Using cache-control setting on htm files in S3 bucket to see if cache-control header of max-age will now be set.


pcraig3 (approved):
🤞

PR #251, opened 2020-09-30
Update content for password reset and 2FA screens

pcraig3:
This PR adds the "V6" content updates for

- the "password reset" pages
- the "2FA" pages

It also does a minor CSS fix so that the 2FA code input is smaller — it's only 6 characters so it doesn't make sense to keep it as a full-sized input.

I also removed the "You have successfully reset your password" page template because we don't use that anymore.

There are some irregularities between our Figma screens and how the app actually works, so I did my best to be consistent.


bryan-robitaille:
Looking good :tada:

bryan-robitaille (approved):
Helps if I press approve instead of comment....

PR #252, opened 2020-09-30
Fix AWS alarm behavior

bryan-robitaille:
AWS metric alarms currently continue to be tripped after an initial alarm because not enough datapoints are available to properly reset the status.  This PR fills missing data with 0 values ensuring that the next evaluated period after the alarm resets is properly evaluated and does not re-trip the alarm based on previous data.  Long winded I know but basically this fixes the behavior when AWS looks further back in time beyond what the time query should be in order to determine if it changes status into alarm.


pcraig3 (approved):
👍

PR #254, opened 2020-10-01
Remove Elastic Beanstalk configuration

pcraig3:
@bryan-robitaille, is there anything else I missed, or is this all the leftover Elastic Beanstalk configuration?

There's also a bunch of screenshots that were accidentally committed to the project root directory (almost certainly by me).
They exist where they need to, so this doesn't break anything — they were just duplicated accidentally.


bryan-robitaille (approved):
👍

PR #255, opened 2020-10-01
Change ID's to be unique for calculated metrics

bryan-robitaille:
Modify calculated metric id's to see if it resolves the new weird behavior in AWS Alarms.


pcraig3 (approved):
🤞

PR #256, opened 2020-10-02
Set default value on Metric Log Filters

bryan-robitaille:
Hoping this will work better than the creation of the calculated metrics in the last PR.  This PR undoes the creation of the calculated metrics and instead replaces it with a pure log filter that includes a default_value that is set to 0.  The previous fix did work when created manually through the AWS interface however did not work as expected when created using Terraform.


pcraig3 (commented):
💯 

Looks more straightforward.

pcraig3 (approved):

PR #257, opened 2020-10-02
Add custom error templates for common error states

pcraig3:
Added error page templates for the following error states:
- 403
- 404
- 500

And we already had one for the CSRF error.

When DEBUG is `True` we can't see these error messages, but going directly to the URL for each one brings up the error.

Helpful sources:
- https://stackoverflow.com/a/37611184
- https://docs.djangoproject.com/en/3.1/topics/http/views/#customizing-error-views

## Screenshots

| before | after |
|--------|-------|
|  white page default error handler      |  looks the part     |
|  Screen Shot 2020-10-02 at 12 07 59 PM   |  Screen Shot 2020-10-02 at 12 08 01 PM  |

For completion, here are all the screens being added.

| error status | screenshot |
|--------|-------|
|   `403`     |   Screen Shot 2020-10-02 at 12 08 09 PM  |
|  `404`      |  Screen Shot 2020-10-02 at 12 08 01 PM  |
|  `500`    |  Screen Shot 2020-10-02 at 12 08 16 PM  |


pcraig3:
Added the revision to the phonetic alphabet mentioned in issue #258

bryan-robitaille (approved):
Very snazzy...
Looking forward to testing these with debug off!

PR #259, opened 2020-10-02
Update the README with our development workflow

pcraig3:
Split into 3 main sections:

- describes about our trunk-based development workflow
- explains how we handle app versioning
- lists the automated tests we run as part of PRs

Also hid the ENV vars in a `<details>` element because it's really crazy how many there are.


bryan-robitaille (approved):
Looking good!

PR #260, opened 2020-10-02
Add throttle check on form submission and post

bryan-robitaille:
This PR fixes a bug where invitations and OTKs could potentially be generated after the throttle limit was hit.

For Invitations:
- On invite page render check to see if self.throttled_active is tripped.  If so render lockout page.

For OTKs:
- On page _post function check to see if self.throttled_active is tripped before trying to get a OTK from the server.  If tripped when page render runs return the lockout page.

### DIFF
--- a/covid_key/views.py
+++ b/covid_key/views.py
@@ -32,7 +32,8 @@ class CodeView(Is2FAMixin, ThrottledMixin, TemplateView):
         token = request.user.api_key
         diagnosis_code = "0000000000"
         covid_key = None
-        if token:
+        self.throttle_limit_check()
+        if token and not self.throttled_active:
             try:
                 try:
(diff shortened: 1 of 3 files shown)


pcraig3 (commented):
Hello! 

Since we're using a [`Mixin`](https://docs.djangoproject.com/en/3.1/topics/class-based-views/mixins/) for this, we shouldn't have to update the view function (ie, `self.throttle_limit_check` shouldn't have to be called in the view).

I figured out that what was happening was that we call "post" first _and then_ we called the ThrottleMixin, so the key was always being generated. If we call the Mixin function first, it should work for both. 

Check out PR #261 for the fix I did.

bryan-robitaille:
Fix introduced in PR #261

PR #262, opened 2020-10-05
Remove record of past login attempts for newly created accounts

pcraig3:
This is a pretty out there edge case, but we've seen it happen.

If someone thinks they have an account and they try to log in a number of times, those attempts will be logged.

If they are then invited to join, they will potentially be locked out of the system because they previously tried to log in without an account.

If someone is being invited to the system, they should never be immediately locked out, so this fixes that behaviour.


bryan-robitaille (approved):
Nice!

PR #264, opened 2020-10-06
Increase the limits for key generation + bump version

pcraig3:
- Increase the limits for AWS' alerts: now needs 40 in an hour to raise a critical alert
- Increase the limit for an individual requesting keys before getting throttled to 100

Also bumps the version so that we can release a new version.


bryan-robitaille (approved):
Let's get this in

PR #273, opened 2020-10-07
Maintenance Page CSS and IMG fixes

bryan-robitaille:
Changing the CSS and Img tag ref's to static instead of relative.  Due to the way S3 and Cloudfront handle routing the original URL does not change when Cloudfront returns a redirect.  For example `/en/login` does not exist on S3/Cloudfront so it returns the default `en.htm` page however the url remains `/en/login`.  This breaks all of the relative paths in the htm file file.  By changing to static it will ensure that the CSS and Img's always render no matter what non-existant URL is passed in.

If there is a better solution I'm all ears!!  👂


pcraig3 (approved):
👍

pcraig3:
This means we'll have to change the relative links on the production site, eh?

bryan-robitaille:
> This means we'll have to change the relative links on the production site, eh?

Yes unfortunately, we'll have to changes the links for the Production specific URL as well.

PR #279, opened 2020-10-08
Add admin_email to context when sending account invitations

pcraig3:
People keep replying to the default notify email (currently creates a ticket in Freshdesk) so we want to mention their portal administrator in the "account invitation" email template.

Note that this only adds a variable to the invitation email template: the Notify template has to be updated to consume this variable. (The variable has to exist in every environment before the template can reference that variable.)

## Emails

| before | after |
|--------|-------|
|   Screen Shot 2020-10-08 at 2 24 03 PM   |  Screen Shot 2020-10-08 at 2 24 13 PM   |


bryan-robitaille (approved):
👍  Looking good!

PR #281, opened 2020-10-09
Convert emails to lowercase during login + bump version

pcraig3:
We are seeing a number of login failures with people entering mixed case email addresses.

We convert emails during signup to lowercase, so we should do it during login as well.

Added a test to make sure that (trailing) whitespace doesn't affect logging in either.

Bump the version so that we can release quickly.


bryan-robitaille (approved):
🎉 👍

PR #287, opened 2020-10-13
Redirect to login if no existing user on yubikey verification page

pcraig3:
We've seen a few 500 errors because of people visiting this page directly without having gone through the login first.

It expects a user object to exist at this point so if there isn't one, we should redirect back to the login page.

Note: no screenshots because this isn't a visual change.


bryan-robitaille (approved):
👍

PR #291, opened 2020-10-14
Fixes to Metric Errors for Security Report

bryan-robitaille:
The original PR that created the metrics for V2 were refused by the Terraform Apply to staging due to 'invalid filter syntax'.  
The WAF for Cloudfront introduced in the same PR was refused during the Terraform Apply because it did not have the 'Global' scope.
This PR fixes these issues.


pcraig3 (approved):
Looks good.

PR #292, opened 2020-10-15
Remove double entry for CloudFront WAF

bryan-robitaille:
This PR removes unnecessary Terraform code that associates the WAFv2 ACL to CloudFront.  The current code is failing due to the nature of the WAFv2 ACL which is GLOBAL instead of REGIONAL.  The WAFv2 ACL is instead associated with the CloudFront site in cloudfront.tf as part of the CloudFront configuration resource.


pcraig3 (approved):
👍  let's merge

PR #295, opened 2020-10-20
Add coverage python library and CI test for 80% test coverage

pcraig3:
This PR:

- adds the [`coverage`](https://coverage.readthedocs.io/en/coverage-5.3/) library to our dev dependencies
- adds `coverage` configuration to `pyproject.toml`
- updates the README with code coverage documentation
- adds some very minor unit tests
- adds a CI code coverage test separate from our regular unit tests 

☝️ that last point is important. Whereas failing the unit tests indicate that something in the app has probably gone wrong, failing the code coverage threshold elevates the risk of publishing the app, but the app probably still works. ie, it is not an explicit indication of unexpected behaviour. Since our 80% coverage limit is unmoored from concrete app performance, we don't want this coverage test to block releases. Unit tests passing is required to merge, but the coverage one is more of a "decide what to do if it happens" kind of limit. 

**Example of a failing coverage test here:** https://github.com/cds-snc/covid-alert-portal/runs/1283363988?check_suite_focus=true

Sources:
- [Django docs on coverage testing](https://docs.djangoproject.com/en/3.1/topics/testing/advanced/#integration-with-coverage-py)
- [`coverage` library](https://coverage.readthedocs.io/en/coverage-5.3/)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,3 +19,9 @@ exclude = '''
                         # the root of the project
 )
 '''
+
+[tool.coverage.run]
+omit = ['*/migrations/*']
+
+[tool.coverage.report]
+fail_under=80.0
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,34 @@
+name: Test code coverage
+
+on: push
+
+jobs:
+  coverage_test:
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 4
+      matrix:
+        python-version: [3.8]
+
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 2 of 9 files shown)


pyproject.toml
```diff
@@ -19,3 +19,9 @@ exclude = '''
                         # the root of the project
 )
 '''
+
+[tool.coverage.run]
+omit = ['*/migrations/*']
```
bryan-robitaille, line 24: Do you also need to exclude the ./config/* folder given that it doesn't include code but the test is running against the root of the folder `./`

bryan-robitaille (approved):
Aprroved but some non critical questions.

PR #296, opened 2020-10-21
Hide "Your account" link on 2FA page

pcraig3:
We had a minor incident where
- someone had changed their phone number
- got the to 2FA page
- couldn't get the text message to their new number
- saw the "Your account" link and clicked it
- wasn't able to see it, ended up on 2fa page again

Basically, you can't see your account until after you've passed the 2FA screen, so we shouldn't have it visible until after people progress past that screen.

### DIFF
--- a/portal/templates/includes/nav_header.html
+++ b/portal/templates/includes/nav_header.html
@@ -3,7 +3,11 @@
 <nav class="nav--header" aria-label="{% trans 'Language and account navigation' %}">
   <ul>
     {% if user.is_authenticated %}
-      <li><a href="{% url 'user_profile' user.id %}">{{ _("Your account") }}</a></li>
+      {# "user.is_authenticated" means they have passed the initial login screen (email, password) #}
+      {# "user.is_verifed" means they have also entered a 2FA code #}
+      {% if user.is_verified %}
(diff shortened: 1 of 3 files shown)


portal/templates/includes/nav_header.html
```diff
@@ -3,7 +3,9 @@
 <nav class="nav--header" aria-label="{% trans 'Language and account navigation' %}">
   <ul>
     {% if user.is_authenticated %}
-      <li><a href="{% url 'user_profile' user.id %}">{{ _("Your account") }}</a></li>
+      {% if user.otp_device %}
```
bryan-robitaille, line 6: Purely for readability I'm wondering if this should instead be `{% if user.is_verified() %} which is also set by Django-OTP.
> OTPMiddleware populates request.user.otp_device to the OTP device object that verified the current user (if any). As a convenience, it also adds user.is_verified() as a counterpart to user.is_authenticated(). It is not possible for a user to be verified without also being authenticated.  

https://django-otp-official.readthedocs.io/en/stable/overview.html#authentication-and-verification

bryan-robitaille (approved):

portal/templates/includes/nav_header.html
pcraig3: yeah that's way better

PR #297, opened 2020-10-21
SLA Dashboard and new Service Alarms

bryan-robitaille:
# Presenting for your approving pleasure... 🥁 

## 2 new Alarms 🚨 :
- 🚦 __Service Availability__ 🚦:  based on the normal rate at which health check requests are logged in the COVID Alert Portal.  If the sum of health check requests received and responded to by the Portal deviates by 2 standard deviations (based on the previous 2 weeks worth of data) then raise a Warning Alarm for further investigation.
- ⏱️  __Response Time__ ⏱️ : based on the average rate of responses within a 60 second period.  If this average response time deviates above 2 standard deviations (based on the previous 2 weeks worth of data) then raise a Warning Alarm for further investigation.
## Service Level Indicator Dashboard 💻  
- ⏱️  __Response Times__ ⏱️ : A time series graph of response times recorded from the Application Load Balancer (The edge of the infrastructure that we control) with an overlaid threshold denoting the 1 second limit.
- ❌  __5xx Error Rate__ ❌ : A calculation that determines the total percentage of 5xx errors generated by the COVID Alert Portal in relation to non 5xx traffic.  This does not include 4xx HTTP errors given the volatility of traffic that in generated by url phishing attacks. 
- 🚦  __Service Availability__ 🚦 : A time service graph of passed health checks with an overlay of expected values.  Any values outside of the expected values are flagged as a service disruption.


config/terraform/aws/cloudwatch.tf
```diff
@@ -236,6 +236,85 @@ resource "aws_cloudwatch_metric_alarm" "invite_lockout_warn" {
...
+  alarm_description   = "COVID Alert Portal Warning - No status checks detected.  Check COVID Alert Portal to ensure site is operational."
+  treat_missing_data  = "notBreaching"
+  alarm_actions       = [aws_sns_topic.alert_warning.arn]
+
+  metric_query {
+    id          = "e1"
+    expression  = "ANOMALY_DETECTION_BAND(m1)"
+    label       = "Status Checks (Expected)"
+    return_data = "true"
+  }
```
pcraig3, line 303: Do we know what 2 standard deviations actually looks like? ie, is it missing 5 status checks?

config/terraform/aws/dashboard.tf
```diff
@@ -223,4 +223,92 @@ resource "aws_cloudwatch_dashboard" "ocio_report" {
...
+        {
+            "type": "metric",
+            "x": 0,
+            "y": 0,
+            "width": 18,
+            "height": 9,
+            "properties": {
+                "metrics": [
+                    [ { "expression": "FILL(m1, 0)", "label": "Reponse Time", "id": "e2" } ],
+                    [ "AWS/ApplicationELB", "TargetResponseTime", "LoadBalancer", "app/covidportal/da875a433e92c979", { "id": "m1", "visible": false } ]
```
pcraig3, line 243: Is `app/covidportal/da875a433e92c979` just the hardcoded id of the load balancer? I guess it would have to change for production.

config/terraform/aws/dashboard.tf
```diff
@@ -223,4 +223,92 @@ resource "aws_cloudwatch_dashboard" "ocio_report" {
...
+            "y": 9,
+            "width": 15,
+            "height": 3,
+            "properties": {
+                "metrics": [
+                    [ { "expression": "SUM([m1,m2])", "id": "e1", "region": "ca-central-1", "yAxis": "left", "label": "5xx HTTP Errors" } ],
+                    [ { "expression": "m4-m3", "id": "e2", "region": "ca-central-1", "yAxis": "left", "label": "Non 4xx HTTP Requests" } ],
+                    [ "AWS/ApplicationELB", "HTTPCode_Target_5XX_Count", "LoadBalancer", "app/covidportal/da875a433e92c979", { "id": "m1", "visible": false } ],
+                    [ { "expression": "(e1/e2)*100", "id": "e3", "region": "ca-central-1", "label": "% 5xx HTTP Error Rate" } ],
+                    [ "AWS/ApplicationELB", "HTTPCode_ELB_503_Count", "LoadBalancer", "app/covidportal/da875a433e92c979", { "id": "m2", "visible": false } ],
```
pcraig3, line 279: What's the difference between `HTTPCode_ELB_503_Count` and regular 500s?

config/terraform/aws/dashboard.tf
```diff
@@ -223,4 +223,92 @@ resource "aws_cloudwatch_dashboard" "ocio_report" {
...
+                    [ ".", "HTTPCode_Target_4XX_Count", ".", ".", { "id": "m3", "visible": false } ],
+                    [ ".", "RequestCount", ".", ".", { "id": "m4", "visible": false } ]
+                ],
+                "view": "singleValue",
+                "region": "ca-central-1",
+                "stat": "Sum",
+                "period": 3600,
+                "stacked": false,
+                "setPeriodToTimeRange": true,
+                "title": "5xx HTTP Errors SLA"
```
pcraig3, line 289: Sorry for being pedandic, but can we call these `SLI`s instead of `SLA`s? 

- SLI: the indicator is something that should tell you something (eg, too many errors, request times are high). We can break our SLIs and nothing happens, but we want to investigate them.
- SLA: you basically are agreeing / guaranteeing to some thresholds with some external group. If these break, you are in trouble.

We don't have agreements in place, we just want to track these numbers.

config/terraform/aws/dashboard.tf
```diff
@@ -223,4 +223,92 @@ resource "aws_cloudwatch_dashboard" "ocio_report" {
...
+                "metrics": [
+                    [ { "expression": "FILL(m1, 0)", "label": "Reponse Time", "id": "e2" } ],
+                    [ "AWS/ApplicationELB", "TargetResponseTime", "LoadBalancer", "app/covidportal/da875a433e92c979", { "id": "m1", "visible": false } ]
+                ],
+                "view": "timeSeries",
+                "stacked": false,
+                "region": "ca-central-1",
+                "period": 60,
+                "stat": "Average",
+                "title": "Response Time SLA",
```
pcraig3, line 250: See other comment, RE: SLAs vs SLIs

pcraig3 (commented):
Just a few comments, have a look through.

config/terraform/aws/cloudwatch.tf
bryan-robitaille: We don't currently know what it looks like because the metric isn't in place.  I can go back though and I'll run an Insight log search to see how many there are in a minute and then calculate the standard deviation.

config/terraform/aws/dashboard.tf
bryan-robitaille: It is the hardcoded load balancer.  Unfortunately I'm not aware of a way to include a relative resource in this type of EOF >> EOF format.

config/terraform/aws/dashboard.tf
bryan-robitaille: The 5xx_Count is coming from the Load Balancer target (aka the portal) and the 503 error is coming from the Loadbalancer.  Basically if the request from the LB to the target fails the LB will return a 503.  So we account for all 5xx errors a client can get by accounting for both origins.

config/terraform/aws/dashboard.tf
bryan-robitaille: Yup, that's a bad catch on my part.  I had switched up SLA for SLI in an earlier commit but missed that one.

config/terraform/aws/dashboard.tf
bryan-robitaille: And this one too apparently.

config/terraform/aws/dashboard.tf
bryan-robitaille: I had swapped out the SLA's for SLI in an earlier commit but it looks like I missed these 2.

config/terraform/aws/cloudwatch.tf
bryan-robitaille: Screen Shot 2020-10-22 at 1 53 44 PM
So you can kinda see how the deviations look like a heart beat?  The machine learning will pick that up and also look for that.  We'll need to tune the number of standard deviations based on what we actually see happening in staging.

bryan-robitaille: So 2 standard deviations from this would be roughly 3.5 health checks per 60s period.

pcraig3 (approved):
I'm approving 🍏

PR #299, opened 2020-10-28
Separate the sign up process into 2 steps

bryan-robitaille:
This PR removes the mobile number from the original registration step and creates and additional page for the mobile number entry.  This is in preparation for the potential of using Static Security Codes.

🎥  🎉  ...............................Get your popcorn ready! .......................................🎉  🎥 
Portal_signup


bryan-robitaille:
Okay, going to update the tests tomorrow to change the assertion now that the flow of the sign up has changed

profiles/forms.py
```diff
@@ -341,21 +341,6 @@ class SignupForm(HealthcareBaseForm, UserCreationForm, forms.ModelForm):
...
-        "required"
-    ]
-
-    phone_number_confirmation = PhoneNumberField(
-        label=_("Confirm your phone number"),
-        help_text=_("Enter the same mobile number as above."),
-    )
-
     password1 = forms.CharField(widget=forms.PasswordInput())
     password1.error_messages["required"] = validation_messages["password"]["required"]
```
pcraig3, line 345: I know it's not part of your PR, but this is something I did by accident in a past PR. If you name these both `password`, instead of `password1`, then it should render as expected.

profiles/signals.py
```diff
@@ -14,7 +14,7 @@ def _save_previous_email(sender, instance, **kwargs):
 
 @receiver(post_save, sender=HealthcareUser)
 def update_2fa(sender, instance, created, **kwargs):
-    if not created:
+    if not created and instance._pre_save_instance.phone_number:
```
pcraig3, line 17: What is this new condition?

profiles/views.py
```diff
@@ -177,6 +177,26 @@ def form_valid(self, form):
...
 
+class SignUp2FAView(LoginRequiredMixin, FormView):
+    form_class = SignupForm2fa
+    template_name = "profiles/signup2fa.html"
+
+    def get_success_url(self):
+        return "{}?next={}".format(reverse_lazy("login-2fa"), reverse_lazy("welcome"))
+
+    def get(self, request, *args, **kwargs):
+        return super().get(request, *args, **kwargs)
```
pcraig3, line 188: The signup form prevents you from seeing it once your token is gone. (eg, once your account is created, you can't revisit the signup URL). I was thinking we could do the same here, but I don't know if it actually makes a difference. This page is equivalent of the /edit/phone-number page basically. 🤷 

What do you think?

pcraig3 (changes requested):
Hello!

Looks good for the most part. I did manage to break the app by trying to visit another URL while still halfway through the signup process (pic below). 

The reason for this is that trying to see the "/start" page, for example, redirects to the 2fa page, and then it looks for the sms token (whoops, no phone == exception!!).

To me it looks like a new `if` check somewhere in the `login-2fa` route, but maybe there's a cleverer way.

Since it was a bug, we should probably also do a test for this.

## error message
Screen Shot 2020-10-29 at 10 13 19 AM

## here's the stack trace
Screen Shot 2020-10-29 at 10 13 31 AM

profiles/forms.py
bryan-robitaille: Will create a new PR for this as it's needed more ugently then this feature

profiles/signals.py
bryan-robitaille: This new condition is needed because the HealthcareUser object is created in the DB on the first signup step with `' '` as the phone number.  With the previous logic when the phone number was entered on the 2nd signup step the function would throw an error because the `old_number` was not a string and tried to perform an `.as_e164` conversion.  It would also fail because the query to find the sms_device would return `None` for the `old_number`.  I played around with various logic combinations and finally landed on this one as the easiest to read.  So new logic is basically:
Is the record already existing and does not have a previous value of `' '` ?  Then find the existing old number in the sms_devices and replace it.  Else just create the new sms_device.

profiles/views.py
bryan-robitaille: Just added a new rule that will bring a user back to the signup-2fa page if they manage to navigate away from the page before inputting a mobile number.  It's essentially an edit/phone-number page but at least it keeps you locked at that page and requires input before you can access the rest of the portal.

pcraig3 (approved):
Looks good!

PR #304, opened 2020-11-02
Anomaly detection Fix

bryan-robitaille:
This PR fixes the two Anomaly detection alarms (ServiceAvailabilityWarn, ResponseTimeWarn) that were learning but not properly comparing incoming data.  It also increases the alarm band to 3 standard deviations which allows for more variance in turn reducing false positives.

This PR also addresses a spelling mistake in the secret "email_host_user"


pcraig3 (approved):
> Quis custodiet ipsos custodes. Bryan est.

translation: "Who detects anomalies in the anomaly detection configuration? It is Bryan".

PR #305, opened 2020-11-03
Anomaly detection - Increase evaluation period

bryan-robitaille:
This PR increases the evaluation period of the anomaly alarms ServiceAvailabilityWarn and ResponseTimeWarn to 5 with a threshold of 2 breaching data points.

Essentially these alarms will now be triggered when their metric dips below the anomaly model's lower expected band at least twice within a 5 minute window.

This table demonstrates this new behaviour.
`0` is a non breaching point
`X` is a breaching point

| Will Trigger Alarm | Will not Trigger Alarm|
| ------------------ | --------------------- |
| 0-X-0-0-X | 0-0-X-0-0 |
| X-X-0-0-0 | 0-0-0-0-0 |


pcraig3 (approved):
Hopefully this stops the constant alarms 🚨 🚨 🚨 🚨

PR #306, opened 2020-11-04
Improvements to the presentation of backup codes + tests

pcraig3:
This PR changes the presentation of the codes on the "generate backup codes" page, after looking through how some other services implemented backup codes. Let's get into it.

A summary of the changes are:
- use a fixed-width font
- visually separate the code into 2 parts (4 chars and 4 chars)
- group them together more clearly in the interface
- put them in an HTML list
- put an "visually hidden" element that capitalizes the characters in the code and inserts spaces between each of the characters

 Later, when backup codes are submitted, we should force lowercase them and remove all whitespace.

## Screenshots

| before | after |
|--------|-------|
|  Screen Shot 2020-11-04 at 11 32 38 AM   |  Screen Shot 2020-11-04 at 11 32 11 AM  |

## Research  

*Note: this is all covered in [this document](https://docs.google.com/document/d/1Ui23pnezBUO-QZVII2zsaysXuz8pNQVHf2hiI4W0MPI/edit#)*.

### Google & Github

A lot of similarities here. Both Google and Github:
- use same-width fonts
- visually separate their codes in the middle
- using semantic HTML for their list of codes (in a 2-column layout) 
- allow you to print or download as a .txt file
- invalidate your old codes when you generate new ones
- let you see existing codes you've generated

Google's codes are 8 characters long, all digits. Github's are 10 characters long, alphanumeric. 

| Google | Github |
|--------|-------|
|   Screen Shot 2020-11-03 at 11 26 02 AM   |   Screen Shot 2020-11-03 at 11 35 30 AM  |

There are problems with the Voiceover implementation for both though. Google has a bunch of blank cells in their table (literally read as "blank"), and in Github's implementation the list items are miscounted and the bullet point in front of the codes is read as "black circle".

### A screen-reader optimized ID code on the Digital Marketplace

Digital Marketplace implementation example

The Digital Marketplace service IDs (eg, [this IBM product](https://www.digitalmarketplace.service.gov.uk/g-cloud/services/808368535412356)) are visually spaced out but when copy+pasted they show up as a continuous string. Also, voiceover reads them out as separate characters. 

id

The HTML implementation of this is one visually hidden ID and then an aria-hidden ID which is separated into blocks of 4.

Screen Shot 2020-11-03 at 4 14 23 PM

So we should separate the presentation of the code from the reading of the code if we want the most control over each format.

### Testing Voiceover

image

Given the following 4 options (note: this was an actual code generated by our algorithm), Voiceover read each one like this ([watch the video](https://drive.google.com/file/d/1uTAZ6NokYGO-3YaYF9fOfjdkgLgJZgKu/view?usp=sharing)):

1. Pronounced all of the letters but read very quickly. The “a” and “e” were hard to distinguish
2. Read the 1st section as letters but tried to read the 2nd section as a word
3. Pronounced all of the letters individually, more clearly then the lowercase letters
4. Tried to read it as a word

Voiceover usually will read the codes out letter-by-letter, but it seems to get caught in some edge cases. For example, if there are no digits in the code AND there are vowels in the code, then it will treat it as a word. The best case of the following options was forcing the letters to uppercase.

### Final implementation 

Using the precedents established by Google and Github's implementations, let's:
- use a fixed-width font
- visually separate the code into 2 parts (4 chars and 4 chars)
- group them together more clearly in the interface
- put them in an HTML list

To improve the screen reader behaviour, let's:
- put an "visually hidden" element that capitalizes the characters in the code and inserts spaces between each of the characters

 Later, when backup codes are submitted, we should force lowercase them and remove all whitespace.

### DIFF
--- a/profiles/static/scss/pages/_2fa-codes.scss
+++ b/profiles/static/scss/pages/_2fa-codes.scss
@@ -1,0 +1,28 @@
+.static-codes {
+  display: inline-block;
+  border: 2px dashed $color-grey;
+  padding: $space-md calc(#{$space-xl} + #{$space-md});
+  margin: $space-md 0 $space-lg 0;
+  border-radius: 10px;
+  text-align: left;
+  position: relative;
+
+  ul {
+    font-family: monospace;
+    font-size: 1.2em;
+    padding-left: 0;
+    margin-bottom: 0;
+    letter-spacing: 0.05em;
+    list-style-type: none;
+
+    li {
+      &:not(:last-of-type) {
+        margin-bottom: $space-sm;
+      }
+
+      [aria-hidden="true"] span:last-of-type {
+        padding-left: $space-xs;
+      }
+    }
+  }
+}
--- a/profiles/static/scss/styles.scss
+++ b/profiles/static/scss/styles.scss
@@ -6,6 +6,7 @@
 @import "./_base";
 
 /* Page-specific styles */
+@import "./pages/_2fa-codes";
 @import "./pages/_login";
 @import "./pages/_number";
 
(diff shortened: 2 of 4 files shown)


bryan-robitaille (approved):
Looks absolutely awesome 👍 🎉

PR #308, opened 2020-11-04
Tune Service Availability Anomaly Alarm

bryan-robitaille:
This PR increases the Standard Deviation on the Anomaly alarm from 3 to 5.  This increase brings the new alarm threshold from 450ms to roughly 700ms.


pcraig3 (approved):
All good!

PR #310, opened 2020-11-04
Separate all the backup/security code stuff into its own django app

pcraig3:
*Note, before merging let's decide on a name to use*

I think we're learning that we don't want to put every bit of functionality in one place. Especially as this set of stories looks like it's getting bigger all the time.

One problem I ran into is that we have like 10 names for these things. Candidates:

- Static codes (this is what the OTP library calls them)
- Security codes (this is what @bryan-robitaille has been calling them)
- Offline codes (this is what the security doc refers to)
- Backup codes (this is what Google calls them)
- Recovery codes (this is what Github calls them)
- Emergency codes 
- One-time codes (?)
- "Scratch cards"

I went with "backup codes" after chatting with Brian, but Brian said he didn't know either and we should wait for a content review. Anyway, happy to merge as is (want to re-write the commit message first) or change after a discussion.

### My opinion
 
I think "backup"/"recovery"/"emergency" all communicate intent. Someone that's not technical can probably tell you what they are for.

"static codes" doesn't mean a whole lot. I don't think "security codes" are very descriptive either. "Scratch cards" is completely out of the question.


backup_codes/views.py
```diff
@@ -0,0 +1,41 @@
...
+from django_otp import DEVICE_ID_SESSION_KEY, devices_for_user
+from django_otp.plugins.otp_static.models import StaticDevice, StaticToken
+
+from waffle.mixins import WaffleSwitchMixin
+
+from portal.mixins import Is2FAMixin
+
+
+class BackupCodeListView(WaffleSwitchMixin, Is2FAMixin, ListView):
+    waffle_switch = "SECURITY_CODE"
```
bryan-robitaille, line 12: We'll probably want to change the Waffle Switch to "BACKUP_CODE" as well

profiles/templates/profiles/healthcareuser_detail.html
```diff
@@ -100,7 +100,7 @@ <h1>{% trans "User profile" %}</h1>
               <td>{{ security_code_count }} {%trans " unused codes remaining" %}</td>
```
bryan-robitaille, line 100: We'll probably want to change the Waffle Switch to "BACKUP_CODE" as well

bryan-robitaille (approved):
I'm fine with backup_codes.   I had gone with security_codes originally because it's what the content in the designs referred to them as.

pcraig3:
Thanks @bryan-robitaille! Gonna merge shortly.

PR #314, opened 2020-11-10
Change policy for Alert Ok

bryan-robitaille:
This PR sets the policy resources (resources the policy can be applied to) to a wildcard.  This is because AWS does not like having multiple resources listed in a single Statement Block.


pcraig3 (approved):
👍

PR #315, opened 2020-11-12
Add a "backup codes" page to the signup

pcraig3:
This PR
- adds a "No mobile phone?" button on the signup screen (behind a flag)
- adds a new "Backup codes" page you can visit if you click the link 
- changes some of the internal url names so that they're always underscores (whereas the url paths users see are always hyphens)
- refactors some CSS in anticipation of the backup codes reusing some of the same CSS we use elsewhere
- adds tests

The most interesting part of this PR is the extra button in the form problem.

<details>
<summary>Read more about the "extra button in the form" problem</summary>
<div>
<br />
Added a "No mobile phone button" and a new screen that generates codes.

Technically this should be a POST request because "GET"ting more codes destroys the previous codes, so it's not just a read. However, it was a bit of a conundrum because the other button was already in a form.

In order to keep the buttons beside each other without resorting to some weird positioning, I just created the second button as link.

Nested forms are bad HTML:

```
<form action="/url1">
<button type="submit">Submit</button>
<form action="/url2"><button type="submit>No? Do this other thing</button></form>
</form>
```

There's apparently a form attribute that you can set on input elements, specifiying the form to activate:
- https://www.sitepoint.com/the-html5-form-attribute/

However, it's not supported in IE11 or lower, and we can't realistically drop support for IE11.

So yeah, made it a GET request -- seemed like the best of bad options.
</div>
</details> 

## Fun gif!

signup


backup_codes/templates/backup_codes/signup_backup_codes_list.html
```diff
@@ -0,0 +1,18 @@
...
+{% block content %}
+  {% include  "includes/flash_messages.html" %}
+  <h1>{% trans "No mobile phone?" %}</h1>
+  <p>{% trans "Log into the portal using one of the security codes below. You can use each code only once.  When you have 1 code left, we’ll tell you how to get more." %}</p>
+  <p>{% trans "Write the codes on paper. Do not save or store this screen, copy the codes to an electronic device, or share the codes with anyone else." %}</p>
+
+  {% include "includes/codes.html" with codes=backup_code_list %}
+
+  <div class="content-button">
+    <a href="#" role='button' draggable='false'>{% trans "Next" %}</a>
```
bryan-robitaille, line 15: This button doesn't seem to perform any action.  After clicking on the button it remains on the same page.

Signup_2fa

bryan-robitaille (changes requested):
I noticed 2 things..
The first is the `Next` button on the `signup_backup_codes_list`.  It doesn't seem to perform any action.
The second is that although the user is created without a phone number associated to them (phone_number is not required on the model) the super admin view doesn't allow the HealthcareUser to be saved without one.  If we're making the phone_number non-required now we should probably ensure both admins and super admins can modify/save a users profile without the `phone_number` filed being required

backup_codes/templates/backup_codes/signup_backup_codes_list.html
pcraig3: Yeah, this is on purpose — I'm waiting for the login story to be completed first (ie, the next screen would be "use a 2fa code to log in").

bryan-robitaille: The 2fa view is the same are the current one.  The existing view accepts both SMS and Backup Codes.  `<a href="{% url 'login_2fa'}" role='button' draggable='false'>{% trans "Next" %}</a>` should do the trick.

pcraig3: Updated!

pcraig3:
I think the phone number problem is mitigated by the new migration. 

> Users and admins will not be able to remove phone numbers once set, but are able to create accounts without initially setting a phone number.

> However, superadmins are able to remove phone numbers using the Django admin backend.

There's a weird account page now for phone numbers if they're empty (it just says "change" — should probably say "add a phone number" or something, but it's still all behind a feature flag so we can fix that later I think.)

pcraig3:
Shoot, I just noticed a problem. You can't get to the 2FA screen without the logic you're adding in your PR.

bryan-robitaille (approved):

pcraig3:
okay, finally merging

PR #316, opened 2020-11-16
Fix logic on token verification.

bryan-robitaille:
Resolving conflicts in a previous Pull Request modified the logic and function signature for the method that verifies a 2fa token.


pcraig3 (approved):
Great catch! ⚾ 

Let's merge when tests pass.

PR #319, opened 2020-11-17
Remove spaces and lowercase submitted 2FA codes

pcraig3:
We display codes in uppercase and with a space in the middle, so that they are easier to copy and remember is the idea.

However, we only accept lowercased codes without spaces in between them.

This PR introduces a method that strips whitespace and lowercases all values submitted through the 2FA form.

## gif

Here's a gif of me logging in with a backup code that I've added uppercase chars to and extra spaces.

Screen Recording 2020-11-17 at 3 56 02 PM


bryan-robitaille (approved):
👍

PR #320, opened 2020-11-18
Security code lockout after 100 Failed attempts

bryan-robitaille:
This PR introduces a new security feature that deactivates an account after 100 bad attempts at either the SMS or Backup Code 2fa.  Upon a successful code entry by a user all 2fa devices will have their failures reset.  This is to ensure that someone with both a SMS device and Backup Codes do not inadvertently deactivate their account if they continuously only use Backup Codes even if they have a SMS device.  Once an account is deactivated the user is automatically logged out and can not login without having the `is_active` flag reset by a Super Admin.

This PR also adds the following tests:
- Test that a user can login with an SMS code.
- Test the lockout feature when using a SMS device
- Test the lockout feature when using Backup Codes
- Test SMS device throttling to ensure their is a delay after bad code entries
- Test Backup Codes throttling to ensure their is a delay after bad code entries


profiles/views.py
```diff
@@ -256,17 +256,21 @@ def form_valid(self, form):
...
-        if not code_verfied and self.has_static_code:
-            code_verfied, backup_being_throttled = verify_user_code(
+        if not locked_out and not code_verfied and self.has_static_code:
+            code_verfied, backup_being_throttled, locked_out = verify_user_code(
                 self.request, code, self.request.user.staticdevice_set.all()
             )
 
+        if locked_out:
+            return redirect(reverse_lazy("login"))
+
```
pcraig3, line 273: We don't have to explicitly log them out? Just setting them to inactive is enough?

backup_codes/views.py
```diff
@@ -175,5 +183,13 @@ def verify_user_code(request, code, devices):
             code_sucessful = True
             request.user.otp_device = device
             request.session[DEVICE_ID_SESSION_KEY] = device.persistent_id
+            reset_all_devices_failure_count(request.user)
 
-    return [code_sucessful, being_throttled]
+    return [code_sucessful, being_throttled, locked_out]
+
+
+def reset_all_devices_failure_count(user):
```
pcraig3, line 191: I should have brought this up earlier, but it's a python convention to use underscore method/variable names when something is a private method (the concept doesn't really exist in Python).

Since in this file, we have some functions that are exported and some that aren't, the idea is that we should prepend an underscore to functions that are only needed in this file vs ones that get used other places in the codebase.

backup_codes/views.py
```diff
@@ -162,8 +160,18 @@ def get_user_backup_codes_count(user):
 def verify_user_code(request, code, devices):
     being_throttled = False
     code_sucessful = False
+    locked_out = False
 
     for device in devices:
+        if device.throttling_failure_count >= settings.BACKUP_CODES_LOCKOUT_LIMIT:
```
pcraig3, line 166: ~~Is this an existing variable or are you setting it right now?~~ nvm, looked it up. https://django-otp-official.readthedocs.io/en/stable/_modules/django_otp/models.html?highlight=throttling_failure_count

profiles/views.py
pcraig3: Looks like it does 👍

pcraig3 (approved):
Looks like it all works pretty well — there's a subtle usabilty type thing where is seems like there's an "off-by-one" error(?). 

eg, I set my "BACKUP_CODES_LOCKOUT_LIMIT" to 3 and then I could submit 3 wrong codes. Only after submitting the 4th code was I returned to the login screen. I get that it's not actually checking the code that 4th time, but it's just something I noticed. 

Gonna approve because I don't really have any substantial comments. Feel free to move functions around if you like, as per my other comments, but also feel free to merge as-is.

👍 🔒

profiles/views.py
```diff
@@ -247,7 +247,7 @@ def get_initial(self):
                 initial["code"] = sms_device.token
             elif self.has_static_code:
                 static_device = self.request.user.staticdevice_set.first()
-                if static_device.token_set.count() > 1:
+                if static_device.token_set.count() > 0:
```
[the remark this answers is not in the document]
bryan-robitaille: 👀

backup_codes/views.py
```diff
@@ -162,8 +160,18 @@ def get_user_backup_codes_count(user):
 def verify_user_code(request, code, devices):
```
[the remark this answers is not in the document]
bryan-robitaille: You're probably right.  Moved it and changed it to a private method naming convention

profiles/views.py
bryan-robitaille: Yep, tested!

backup_codes/views.py
bryan-robitaille: Refactored the functions and moved some that were only used in profiles.view and renamed to private methods.

PR #321, opened 2020-11-20
Admins getting security codes for other users 

pcraig3:
## Summary

This PR builds out the flow whereby an admin can create codes for other users. 
- At any point, an admin can generate 1 code for any staff users or any other admin users in their province. 
- Doing so will replace any other codes that user has previously generated. 
- Admins _can't_ create codes for users in other provinces or superusers.

This PR also:
- entends the phonetic alphabet so that it works for all letters of the alphabet (checked with Yedida already about this)
- moves some shared code to the `portal` so that it can be shared between modules
  - specifically, the `ProvinceAdminMixin` and the `alphabet` template tag
- lightly refactors the `profiles/tests.py` file to remove some unneeded lines — no functional changes

Easiest to show by way of a gif.

## get 'cher gifs 'ere

Screen Recording 2020-11-20 at 12 07 46 PM


bryan-robitaille (approved):
Looks awesome!

PR #322, opened 2020-11-23
Add print styles for backup_codes pages

pcraig3:
## Summary

Added print styles only to the pages that a user would see when getting a backup code for themselves (not going to add it to the admin one).
For all other pages of the site, printing will keep all the visible elements, including the header and footers.

For the backup codes pages, the print styles:
- hide the header, footer, Canada FIP elements, the on-page buttons, and the navs
- show a COVID alert logo image
- show an alternative title with the email address of the logged-in user
- bump up the font sizes for everything

Didn't use different text than what was on the screen, as that would have been a lot more logic than I wanted to do.

## Print preview 

| before | after |
|--------|-------|
|  canada header, nav elements, button at the bottom, wants to print 2 pages      |  1 page, logo in header, user email in title, bigger fonts      |
|   Screen Shot 2020-11-23 at 4 46 11 PM  |  Screen Shot 2020-11-23 at 4 46 36 PM  |


bryan-robitaille (approved):
Tested on Heroku deploy.
Looking good! 👀  👍

PR #328, opened 2020-11-30
Add signal to remove pending announcement creation on user deletion

bryan-robitaille:
This PR ensure that when new announcements are created during a HealthcareUser's cascade deletion that those new announcements are also deleted before the DB update is done.


pcraig3 (changes requested):
Can you write a test for this? 

(Hint, there's one over here you can use: https://github.com/cds-snc/covid-alert-portal/commit/7171c0a88acc4ad8d21b00b0617d72c0e735b937#diff-0e7f46abe26fa8a6b8c0be01a304f2bf3d6810f3092b97b4765fb406755cd8d4)

pcraig3 (approved):
👍  nice! hopefully all bugs we can squash them before they go to prod.

bryan-robitaille:
Test now included

pcraig3 (changes requested):
Wait!

bryan-robitaille:
Ok

pcraig3 (changes requested):
Since this isn't in the `profiles` module, we should use `get_user_model` instead of hardcoding the HealthcareUser.

eg,

```py
from django.contrib.auth import get_user_model
User = get_user_model()
```

pcraig3 (approved):
once tests are green, let's merge.

PR #329, opened 2020-11-30
Add DNS healthcheck alarm to root site URL

bryan-robitaille:
This PR adds a healthcheck to the route domain `staging.covid-hcportal.cdssandbox.xyz` to ensure it is routeable and creates an alarm if ever the site becomes unroutable from outside our AWS hosting.


config/terraform/aws/route53.tf
```diff
@@ -50,3 +50,13 @@ resource "aws_route53_record" "covidportal_maintenance" {
...
 
+# Route 53 Healthcheck
+resource "aws_route53_health_check" "dns-is-routeable" {
+  fqdn              = "staging.covid-hcportal.cdssandbox.xyz"
+  port              = 443
+  type              = "HTTPS"
+  resource_path     = "/status/"
+  failure_threshold = "5"
+  request_interval  = "30"
+
```
pcraig3, line 61: Does this mean a failure counts as 5 checks that fail (and checks happen every 30 seconds)?

config/terraform/aws/cloudwatch.tf
```diff
@@ -328,6 +328,27 @@ resource "aws_cloudwatch_metric_alarm" "service_availability_critical" {
...
 
+###
+# AWS Cloudwatch Alarms - DNS routing error
+###
+resource "aws_cloudwatch_metric_alarm" "service_dns_critical" {
+  alarm_name          = "ServiceDNSCritical"
+  comparison_operator = "LessThanThreshold"
+  evaluation_periods  = "1"
+  threshold           = "1"
+  period              = "60"
```
pcraig3, line 339: This means that if the health check fails 1 time, it does an alarm right? Does this automatically notify Slack, OpsGenie, etc??

config/terraform/aws/route53.tf
bryan-robitaille: Yes, if the DNS check fails for a period  of 2.5 min switch to unhealthy

config/terraform/aws/cloudwatch.tf
bryan-robitaille: Yes the moment that the DNS check fails 1 time (so technically already failing for 2.5 minutes) raise the alarm.  This is staging so we're not raising OpsGenie but this will send a message through slack because it leverages the Warning SNS instance.
```
  alarm_actions       = [aws_sns_topic.alert_warning.arn]
  ok_actions          = [aws_sns_topic.alert_ok.arn]
```

pcraig3 (approved):
Approved!

PR #330, opened 2020-11-30
Update account page for no phone number or codes

pcraig3:
# Summary

This PR:
- Updates the text in front of the security codes on the signup page and the "list" page
- Updates the quick guide to mention users without phones
- Updates the "account" page to show if you have not entered a phone number or generated codes yet
- Fixes the visually-hidden CSS

Most of the updates are obvious but the CSS one I'll explain a bit

### Fixes the visually-hidden CSS

I had been using a snippet of GOV.UK styling but I was finding that on the "codes" page, I was able to select 1 invisible character from the "visually-hidden" span, and then the code wouldn't work.

Using the bootstrap CSS seems to fix this problem.

## Screenshots

| before | after |
|--------|-------|
|   "change" empty mobile phone number, 0 codes remaining     |  "Add" mobile phone number, No codes generated     |
|  Screen Shot 2020-11-30 at 4 30 10 PM  |   Screen Shot 2020-11-30 at 4 30 14 PM |

| before | after |
|--------|-------|
|        |   this section is updated across a few different pages (signup, account, print page)    |
|  Screen Shot 2020-11-30 at 4 31 29 PM     |  Screen Shot 2020-11-30 at 4 31 31 PM   |


bryan-robitaille (approved):
👍

PR #333, opened 2020-12-01
Dns Alarm - Add SNS in US-East-1

bryan-robitaille:
This PR adds a new SNS (Notification System) in US-East-1 as Cloudwatch alarms cannot communicate across regions.  This SNS is only used for the Route53 DNS Alarm.


pcraig3 (approved):
No more okay action?

bryan-robitaille:
> No more okay action?

The okay action lives in Canada-Central... we could create one in US-East if that's something you'd like.

bryan-robitaille:
Added 'Everything is ok' alert for DNS returning to normal

PR #334, opened 2020-12-02
GA Tag inserter module

bryan-robitaille:
This PR adds another module in django that allows for the configuration of the Google Analytics V4 tag from the admin interface.  This was done to ensure that only the production site had the proper Google Analytics property ID setup.  The Content Security Policy was also updated to include the usage of a nonce to enhance security for injected scripts (aka Google Analytics).


google_analytics/templates/includes/google_analytics.html
```diff
@@ -0,0 +1,17 @@
+{% load solo_tags %}
+
+{% block content %}
+  {% get_solo 'google_analytics.GoogleAnalytics' as ga_config %}
+  {% if ga_config.ga_code %}
+    <!-- Global site tag (gtag.js) - Google Analytics -->
+    <script nonce={{request.csp_nonce}} async src="https://www.googletagmanager.com/gtag/js?id={{ga_config.ga_code}}"></script>
+    <script nonce={{request.csp_nonce}}>
```
pcraig3, line 8: Not sure if this needs quotes or not, but I would tend to add them.  (eg, `nonce="{{request.csp_nonce}}"`.

Interestingly, I thought it wasn't coming through at all but [you're not supposed to see the `nonce` value in the DOM explorer](https://stackoverflow.com/questions/55670985/google-chrome-stripping-nonce-values-from-script-tags).

Learn something new every day 😉

google_analytics/templates/includes/google_analytics.html
```diff
@@ -0,0 +1,17 @@
+{% load solo_tags %}
+
+{% block content %}
+  {% get_solo 'google_analytics.GoogleAnalytics' as ga_config %}
+  {% if ga_config.ga_code %}
+    <!-- Global site tag (gtag.js) - Google Analytics -->
```
pcraig3, line 6: We can get rid of either this one or the one in `base.html`. When rendered, there are 2 comments on top of each other, each saying the same thing.

image

pcraig3 (approved):
I mean, all looks good. I made 2 small comments so look at those before you merge.

google_analytics/templates/includes/google_analytics.html
bryan-robitaille: Doesn't need the quotes.  The value when printed straight to the DOM actually contains the quotes.

google_analytics/templates/includes/google_analytics.html
bryan-robitaille: Done :)

bryan-robitaille: Remove the comment from the partial.

pcraig3:
Looks good!

pcraig3 (approved):

PR #335, opened 2020-12-04
Remove BACKUP_CODE feature flag now backup codes are live

pcraig3:
🎉

old code is bunk


pcraig3:
Fixed the extra import ☝️

bryan-robitaille (approved):
🎉   🔔  Ding dong the feature flag is gone 🔔  🎉

PR #336, opened 2020-12-09
Reference SCSS from announcements module in the portal module

pcraig3:
As I started building the new pages, I noticed what a faff it is to have all the CSS in the `/profiles` subfolder, but it's also a faff to have to compile the announcements module CSS separately from the rest of the app.

Went into the [`django-sass`](https://pypi.org/project/django-sass/) package to see what their recommendation and they seem to be recommending that you link your dependencies between modules:

```
@import 'app1/scss/colors';
@import 'app2/scss/colors';
// Or since you are in app2, you can reference its colors with a relative path.
@import 'colors';
```

Basically, we have a few options here but I think that it's fair to make announcements less independent if it makes our code more comprehensible (and solves the problem of adding CSS in other modules).

## Options

1. Keep it as is — modules are independent
2. Link CSS for all modules from `profiles` (eg, here we are linking `announcements` CSS)
3. Move most CSS to `portal` instead of `profiles` (bigger refactor, requires changes to deployment scripts etc).

My fave option is 2, even though it's technically the least correct. From a practical perspective, we're not planning to release any of our modules on pypi or whatever, so option 2 just makes it easier to manage CSS for this project.

Sources: 
- https://pypi.org/project/django-sass/


bryan-robitaille (approved):
I'm good with that.  Makes it easier to manage the scss.

PR #338, opened 2020-12-11
Make main nav visible for not-logged-in users + bump version

pcraig3:
# Summary

Means they can see the "about" page as well as the "support" page before logged in (or if they can't log in).

Also bumps the version so that we can release the change.

## Screenshots

| before | after |
|--------|-------|
|   thin blue line      |  #acab      |
|  Screen Shot 2020-12-11 at 3 55 42 PM   |  Screen Shot 2020-12-11 at 3 55 40 PM   |


bryan-robitaille (approved):
Nice!!

PR #344, opened 2020-12-16
Remove h3s, smaller h2s

pcraig3:
# Summary

This PR 
- removes the h3s we were using on the about pages
- removes the "next" link to the support page
- resizes some of the headings
 
I spent a lot more time on this than the number of commits would imply. Basically, I was on [type-scale.com](https://type-scale.com/) playing with different values and trying them out. In the end I went with [Minor Third at a base size of 20.8px](https://type-scale.com/?size=20.8&scale=1.200&text=A%20Visual%20Type%20Scale&font=Poppins&fontweight=400&bodyfont=body_font_default&bodyfontweight=400&lineheight=1.75&backgroundcolor=%23ffffff&fontcolor=%23000000&preview=false).

Since we don't really want to use h3s, I focused on the h1s and h2s — basically I was looking for a subtle change rather than a jarring one. In the end, the h1 is a tiny bit larger, the h2s are a lot smaller and h3s will be very small. 

## Screenshots

Used the Support page for the screenshot because it has a lot of text on it.

Before, we had h1s, h2s, h3s, leading to lots of font sizes. Now we just have h1s and h2s, but they are closer to the text size (although larger than the old h3s were). I think it looks better, but I spent too long on this story and maybe it looks the same.

| before | after |
|--------|-------|
| healthcare-portal herokuapp com_en_support_      |   127 0 0 1_8000_en_support_ (1))   |

### DIFF
--- a/about/templates/about/3_one_time_keys.html
+++ b/about/templates/about/3_one_time_keys.html
@@ -18,7 +18,7 @@
     <h2>{% trans "Before generating a key" %}</h2>
     <p>{% trans "You’ll confirm that the patient should get a key, and that they’re ready to enter or record it. The portal will guide you each time you need to do this." %}</p>
 
-    <h3>{% trans "Only generate a key if you intend to give it to a patient" %}</h3>
+    <h2>{% trans "Only generate a key if you intend to give it to a patient" %}</h2>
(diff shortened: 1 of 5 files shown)


bryan-robitaille (approved):
LGTM 👍

PR #347, opened 2020-12-21
Changes to "Change your password" flow

### LINKED ISSUE #340: Changes to "Change your password" flow
@pcraig3 

Please change as follows in both English and French
All the hint text is on the screen, but there should be just one line: “Your password must be at least 12 characters and hard to guess.” 
The other hint text is only meant to show if the user makes that particular error.

Please replace 
"Confirm your password
Enter the same password as above."
 with “Enter your password again”   and in French with "Saisissez de nouveau votre mot de passe"

PierrePaul:
This PR fixes #340 issue, removes some help text from PasswordEdit form, depending on context.
This also fixes the validation not being performed on the actual user making the request.
I also added the `email` to the `HealthcareUserAttributeSimilarityValidator`, to prevent people from using `super@email.com123` as password. Most of the time emails contains the name of the person (which is already validated), but just in case.

- Login with your account
- Top right, Manage your account
- Next to `Password`, click `Change`

# Unresolved questions / Out of scope | Questions non résolues ou hors sujet

Maybe we should disable all help text when there is an error. This PR only hides the password_too_short help_text if the password_too_short is in the error list, to prevent a screen reader from reading it two times.

# Reviewer checklist | Liste de vérification du réviseur

This is a suggested checklist of questions reviewers might ask during their review | Voici une suggestion de liste de vérification comprenant des questions que les réviseurs pourraient poser pendant leur examen :

- [ ] Does this meet a user need? | Est-ce que ça répond à un besoin utilisateur?
- [ ] Is it accessible? | Est-ce que c’est accessible?
- [ ] Is it translated between both offical languages? | Est-ce dans les deux langues officielles?
- [ ] Is the code maintainable? | Est-ce que le code peut être maintenu?
- [ ] Have you tested it? | L’avez-vous testé?
- [ ] Are there automated tests? | Y a-t-il des tests automatisés?
- [ ] Does this cause automated test coverage to drop? | Est-ce que ça entraîne une baisse de la quantité de code couvert par les tests automatisés?
- [ ] Does this break existing functionality? | Est-ce que ça brise une fonctionnalité existante?
- [ ] Should this be split into smaller PRs to decrease change risk? | Est-ce que ça devrait être divisé en de plus petites demandes de tirage (« pull requests ») afin de réduire le risque lié aux modifications?
- [ ] Does this change the privacy policy? | Est-ce que ça entraîne une modification de la politique de confidentialité?
- [ ] Does this introduce any security concerns? | Est-ce que ça introduit des préoccupations liées à la sécurité?
- [ ] Does this significantly alter performance? | Est-ce que ça modifie de façon importante la performance?
- [ ] What is the risk level of using added dependencies? | Quel est le degré de risque d’utiliser des dépendances ajoutées?
- [ ] Should any documentation be updated as a result of this? (i.e. README setup, etc.) | Faudra-t-il mettre à jour la documentation à la suite de ce changement (fichier README, etc.)?


portal/settings.py
```diff
@@ -166,15 +166,15 @@
...
 AUTH_PASSWORD_VALIDATORS = [
     {
         "NAME": "profiles.validators.HealthcareMinimumLengthValidator",
-        "OPTIONS": {"min_length": 12},
+        "OPTIONS": {"min_length": PASSWORD_MINIMUM_LENGTH},
     },
     {
         "NAME": "profiles.validators.HealthcareUserAttributeSimilarityValidator",
-        "OPTIONS": {"user_attributes": ["name", "username"]},
+        "OPTIONS": {"user_attributes": ["name", "username", "email"]},
```
pcraig3, line 177: lol, this is a good catch. since the login uses a "username" field as the email, I figured that they were equivalent. Good to do this instead though.

profiles/forms.py
```diff
@@ -199,15 +199,17 @@ class HealthcarePasswordEditForm(HealthcareBaseEditForm):
         label=_("Password"),
         strip=False,
         widget=forms.PasswordInput(),
-        help_text=password_validation.password_validators_help_text_html(),
+        help_text=_(
+            f"Your password must be at least {settings.PASSWORD_MINIMUM_LENGTH} characters."
+        ),
```
pcraig3, line 204: Hmm, I want to keep the messages there. [As I said in the issue](https://github.com/cds-snc/covid-alert-portal/issues/340#issuecomment-752692578), I think it's helpful to have more information there, even if it's redundant.

profiles/forms.py
```diff
@@ -199,15 +199,17 @@ class HealthcarePasswordEditForm(HealthcareBaseEditForm):
...
     )
     password1.error_messages["required"] = validation_messages["password"]["required"]
 
     password2 = forms.CharField(
-        label=_("Confirm your password"),
+        label=_("Enter your password again."),
         widget=forms.PasswordInput(),
         strip=False,
-        help_text=_("Enter the same password as above."),
+        help_text=None,
```
pcraig3, line 212: This is good though, we don't need the second hint text.

profiles/forms.py
```diff
@@ -237,8 +239,14 @@ def _post_clean(self):
...
             try:
-                password_validation.validate_password(password, self.instance)
+                password_validation.validate_password(password, self.user)
             except ValidationError as error:
+                for error in error.error_list:
+                    if error.code == "password_too_short":
+                        # Let's remove the help text to prevent displaying
+                        # the same message twice, we might argue we should
+                        # always remove the help_text when there is an error
+                        self.fields["password1"].help_text = None
```
pcraig3, line 249: Let's remove this. I think let's just leave the hint text and validation messages in place on the first input.

profiles/forms.py
```diff
@@ -199,15 +199,17 @@ class HealthcarePasswordEditForm(HealthcareBaseEditForm):
...
-        help_text=password_validation.password_validators_help_text_html(),
+        help_text=_(
+            f"Your password must be at least {settings.PASSWORD_MINIMUM_LENGTH} characters."
+        ),
     )
     password1.error_messages["required"] = validation_messages["password"]["required"]
 
     password2 = forms.CharField(
-        label=_("Confirm your password"),
+        label=_("Enter your password again."),
```
pcraig3, line 209: This one needs a French translation, I think.

pcraig3 (changes requested):
Hello! Thanks for the contribution! I was off last week but back this week, so I am just getting to this. Two things.

1. Very small changes 

Basically, I think let's do reversions to the hint text stuff for the first field. And the one changed line needs a french translation.

2. Snyk / Unit tests fail

Not really sure what to do about this, but I don't really want to spend too long on this for such a small and obviously useful change. If you update the branch with the changes, I can fiddle with the settings and get it to be mergeable. 

What do you think? Does that make sense?

profiles/forms.py
PierrePaul: You're right. Removing `#fuzzy` on the new translation, keeping the previous translation though : `Saisissez votre mot de passe`

PierrePaul:
Thanks for the review!
Updated the PR with the requested changes.

For Snyk, I just can't run it, it gives me this error : `snyk requires an authenticated account. Please run 'snyk auth' and try again.`

pcraig3 (approved):
Approved! Thanks for the contribution!

PR #348, opened 2020-12-31
Collapsing nav

pcraig3:
# Summary 

This PR brings in 2 changes:

- Add accordion-like "side" nav for mobile-sized about pages
  - The side nav we have currently looks pretty good on a desktop browser, but on mobile it is basically just in the way
- Prevent visually-hidden text from being selectable

## Test instructions

- Open up the Heroku PR app 
- Click any of the about pages
- Resize your browser

## gif

Here's how it looks.

### DIFF
--- a/about/templates/includes/nav_about.html
+++ b/about/templates/includes/nav_about.html
@@ -1,14 +1,13 @@
+{% if request.resolver_match.url_name == 'start' %}class="active"{% endif %}
 {% load i18n %}
 
-{% if request.resolver_match.url_name == 'start' %}class="active"{% endif %}
+<div class="nav--about__desktop">
+  {% include  "includes/_nav_about.html" %}
+</div>
 
-<nav class="nav--about" aria-label="{% trans 'Table of contents: About the Portal' %}">
-  <ul>
-    <li><a {% if request.resolver_match.url_name == 'index' %}class="active"{% endif %} href="{% url 'about:index' %}">{% trans "COVID Alert" %}</a></li>
-    <li><a {% if request.resolver_match.url_name == 'create_account' %}class="active"{% endif %} href="{% url 'about:create_account' %}">{% trans "Create an account" %}</a></li>
-    <li><a {% if request.resolver_match.url_name == 'one_time_keys' %}class="active"{% endif %} href="{% url 'about:one_time_keys' %}">{% trans "One-time keys" %}</a></li>
-    <li><a {% if request.resolver_match.url_name == 'give_a_key' %}class="active"{% endif %} href="{% url 'about:give_a_key' %}">{% trans "Give a key" %}</a></li>
-    <li><a {% if request.resolver_match.url_name == 'help_patient_enter_key' %}class="active"{% endif %} href="{% url 'about:help_patient_enter_key' %}">{% trans "Help patient enter key" %}</a></li>
-    <li><a {% if request.resolver_match.url_name == 'admin_accounts' %}class="active"{% endif %} href="{% url 'about:admin_accounts' %}">{% trans "Admin accounts" %}</a></li>
-  </ul>
-</nav>
+<details class="nav--about__mobile">
+  <summary>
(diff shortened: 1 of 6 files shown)


pcraig3:
Hey @amazingphilippe, I've tried with two different colours and it's hard to choose between them.

| lighter | darker |
|--------|-------|
|  Screen Recording 2021-01-04 at 10 45 11 AM   |  Screen Recording 2021-01-04 at 10 46 14 AM  |

amazingphilippe:
Ah! We could flip a coin!

I lean towards the darker because it feels more noticeable. And the ultra light shades of gray sometimes don’t display well on some poorly calibrated screens. 

Is it me or does the blue also change? We should check if they at least have 7:1 contrast on hover. 

Also wanted to ask if the menu toggle card for the actual nav or only the sidebar?

Philippe Caron
Canadian Digital Service (CDS) | Service numérique canadien (SNC)
Treasury Board of Canada Secretariat | Sécretariat du Conseil du Trésor du Canada
philippe.caron@cds-snc.ca | 343-542-7841

N'hésitez pas à répondre dans la langue officielle de votre choix.
Please do not hesitate to reply in the official language of your choice.

pcraig3:
Okay! Updated to use "About the Portal" instead of "Topics", so the french translation is already there.

bryan-robitaille (approved):
Looking awesome!! 🎉

amazingphilippe:
LGTM as well. For archival purposes :)

about/templates/includes/nav_about.html
```diff
@@ -1,14 +1,13 @@
...
-    <li><a {% if request.resolver_match.url_name == 'give_a_key' %}class="active"{% endif %} href="{% url 'about:give_a_key' %}">{% trans "Give a key" %}</a></li>
-    <li><a {% if request.resolver_match.url_name == 'help_patient_enter_key' %}class="active"{% endif %} href="{% url 'about:help_patient_enter_key' %}">{% trans "Help patient enter key" %}</a></li>
-    <li><a {% if request.resolver_match.url_name == 'admin_accounts' %}class="active"{% endif %} href="{% url 'about:admin_accounts' %}">{% trans "Admin accounts" %}</a></li>
-  </ul>
-</nav>
+<details class="nav--about__mobile">
+  <summary>
+    <div>{% trans "About the portal" %}</div>
+  </summary>
+    {% include  "includes/_nav_about.html" %}
```
andrej-vasilj, line 12: I haven't seen a solution like this before where you inject the entire navbar html two times, once for desktop and once for mobile. I see that it does have the advantage of not requiring any corresponding JS which is nice. I wonder though if it would make sense to just include the details element in desktop mode as well, boot it as 'open' and hide the summary button? The downside would be that the menu starts in the open state for mobile as well...

andrej-vasilj (approved):
Looks good to me. Just had one question which I left as an inline comment.

about/templates/includes/nav_about.html
pcraig3: Yeah, good question. I (we) try to avoid using JS when we don't need to — eg, using the details element at all is a way to do this. 

But! as you rightly point out, now we have 2 menus in the markup. If focus handling were important (eg, you can lose your focus if you change your screen size), then this could be complicated, but it's not really.  

I considered what you are suggesting, but I wasn't able to square the circle without JS. We could use CSS to hide the "summary" heading on desktop mode, but then the menu toggle has to start in "open" mode, which defeats the purpose of having a collapsible menu, if it always starts in open mode.  We could use JS to listen for screen sizes to decide on initial load (or when the screen size is adjusted) whether to have it open or not, but I'm happy to not do that and just repeat the menu. 

It looks a bit hack but it's valid markup and users won't notice, so that's where I ended up.

andrej-vasilj: Ya that's the same conclusion that I came to as well, but I'm happy to hear that you considered the alternatives. Thanks for the clarifications, makes sense to me!

PR #351, opened 2021-01-06
Added a confirmation email upon user signup

andrej-vasilj:
# Summary | Résumé

Trello link: https://trello.com/c/HJkzDHtm

The PR adds code to automatically send a confirmation email once a user has successfully completed the signup process

1. Use an existing admin user to create an invitation for a new user (enter an accessible email address)
2. Once the invite has been received, use it to follow the user signup link within the email
3. After completing the signup, verify that you have received a new "confirmation" email


config/terraform/aws/ecs.tf
```diff
@@ -42,8 +42,10 @@ data "template_file" "covidportal_task" {
     invitation_email_template_id_fr        = aws_secretsmanager_secret_version.invitation_email_template_id_fr.arn
     password_reset_email_template_id_en    = aws_secretsmanager_secret_version.password_reset_email_template_id_en.arn
     password_reset_email_template_id_fr    = aws_secretsmanager_secret_version.password_reset_email_template_id_fr.arn
-    backup_code_admin_email_template_id_en = aws_secretsmanager_secret_version.backup_code_admin_email_template_id_en.arn,
-    backup_code_admin_email_template_id_fr = aws_secretsmanager_secret_version.backup_code_admin_email_template_id_fr.arn,
```
pcraig3, line 46: Interesting that those last two values had commas -- looks like a code linting problem with someone's editor

profiles/views.py
```diff
@@ -177,6 +182,13 @@ def form_valid(self, form):
 
         return super(SignUpView, self).form_valid(form)
 
+    def get_inviter(self):
+        # get invite object and the admin user who sent the invite
+        invited_email = self.request.session.get("account_verified_email", None)
+        invitation = Invitation.objects.get(email__iexact=invited_email)
+        admin = HealthcareUser.objects.get(id=invitation.inviter_id)
+        return admin
```
pcraig3, line 190: This is really the only thing I'm seeing that it would be easy to write a test for. You would have to do bit of setup, but you could get one going I think.

pcraig3 (commented):
Heyo! A few things, neither of which are that critical. 

1. If you could write a test for `get_inviter`, that would be good.
2. Looks like CI is failing because of `black` formatting. If you run `pipenv run black`, it will reformat whatever file it's unhappy with.
3. There's a changelog conflict. Can you update the changelog + bump the version number? Then we can do a release once this is merged.

portal/.env.example
```diff
@@ -9,10 +9,14 @@ API_AUTHORIZATION = ADD_TOKEN_HERE
...
 # DATABASE_URL =
 
+# BACKUP_CODE_ADMIN_EMAIL_TEMPLATE_ID_EN =
+# BACKUP_CODE_ADMIN_EMAIL_TEMPLATE_ID_FR =
 # INVITATION_EMAIL_TEMPLATE_ID_EN =
 # INVITATION_EMAIL_TEMPLATE_ID_FR =
 # PASSWORD_RESET_EMAIL_TEMPLATE_ID_FR =
 # PASSWORD_RESET_EMAIL_TEMPLATE_ID_EN =
+# CONFIRMATION_EMAIL_TEMPLATE_ID_EN =
+# CONFIRMATION_EMAIL_TEMPLATE_ID_FR =
```
pcraig3, line 19: Let's add this to the README as well

pcraig3 (approved):
So I was initially skeptical — new abstractions, more overhead, personally I am not familiar with this library, etc, all the usually things.

But actually it's turned out pretty clean and, you're right, as we keep scaling the Notify API client, it scales less and less. In the end, using the injector is a pretty good solution and cleans up our code base a bit (you can see it in the tests). 

Neat work 🎉 .

Let's update the README and the VERSION before merging, and then we can test it all works on staging.

andrej-vasilj:
@pcraig3 I understood that if we update the verion it will deploy to production automatically?

PR #354, opened 2021-01-08
Create two lists for backup codes

pcraig3:
# Summary

Instead of creating one list for backup codes, where each list item has a visually hidden code and a visually visible code, let's have 2 lists. One list is hidden, and one list is visible.

## Context

The problem with the last approach was that selecting all of the codes at once would duplicate the codes: the visually-hidden codes were still being selected. 

The `user-select` CSS rule doesn't let you select the text directly, but if you selected the text before and after it, the intermediate text is selected regardless. You can see a demo of that here: https://codepen.io/chriscoyier/pen/ZGKYxj

Tried a bunch of CSS and a bunch of JS and it was just easier to have 2 lists and have that be the end of it. No visual changes for users, but copying and pasting will work better.

Note that it doesn't work perfectly. You can still select both sets of codes if you select some text from the paragraph beforehand, and then start selecting backup codes. What I would say to that is that this is still an improvement and I am open to suggestions.

1. Log in to the review app
2. Go to your profile and generate backup codes
3. Try copying a backup codes / multiple backup codes into another document
4. Make sure you only see 1 set of codes

### Unresolved questions / Out of scope | Questions non résolues ou hors sujet

If there's a better way to do this, pls let me know.

## gifs

| before | after |
|--------|-------|
|  Screen Recording 2021-01-08 at 4 48 21 PM    |   Screen Recording 2021-01-08 at 4 46 57 PM   |


andrej-vasilj (approved):
LGTM

Just for my own understanding, why do we need the hidden list of codes? I guess it's some sort of accessibility feature?

Maybe it would be worth considering having a "copy codes to clipboard" button as well. I think this is the typical design pattern on most websites these days. This way you can decide exactly what gets copied.

pcraig3:
> Just for my own understanding, why do we need the hidden list of codes? I guess it's some sort of accessibility feature?

Yup, exactly. If you use a screen reader for the 8-character codes, it sometimes reads it as a word, rather than reads out each character individually. I'm writing a blog post about it, so it will probably get a million shares once I do.

> Maybe it would be worth considering having a "copy codes to clipboard" button as well. I think this is the typical design pattern on most websites these days. This way you can decide exactly what gets copied.

Yeah, I think we talked about something like that. @amazingphilippe may or may not be looking into it. 

We don't have a high takeup rate on the codes though, so maybe we leave it for now.

PR #355, opened 2021-01-12
Add explicit version numbers to dependencies

pcraig3:
Without explicit version numbers, we might don't really have a sense of when packages change, and we might end up updating a package that has a breaking change.

Version numbers are the most recent of all of our packages.


andrej-vasilj (commented):
I'm not sure if changing the pipfile itself is actually necessary or desirable. This is what the lockfile is for and the dependencies get installed using the lockfile which will already give us package version stability as far as I understand. Am I missing something here?

pcraig3:
I don't actually have a super strong opinion on this but I would tend to agree with you. It's true we have the lockfile, but because versions aren't explicit in the Pipfile, we can easily update them without being too conscious of it (eg, I don't really look at updates to the lockfile.

Basically it comes down to "should we have super explicit version numbers" or "should we keep doing upgrades all the time".  

@dsamojlenko is moving onto this team so we can ask his opinion, but if we all don't think this is that important than happy to close this PR.

pcraig3:
Yeah, @andrej-vasilj, you're right about this. Closing this PR.

PR #357, opened 2021-01-13
OTK SMS spike working

andrej-vasilj:
# Summary | Résumé

This work is related to sending the one-time-key (OTK) to patients via SMS. When a new key has been generated, portal users will have the option to navigate to a new form that allows them to enter a patient's mobile phone number and send them the generated key via SMS.

To enable this feature, it was necessary to cache the generated key within the server-side session. Currently, when the user navigates back to the start page, the cache is deleted.

1. Enable SMS for the province of your user (go to admin panel and edit the provinces to enable 'sms_enabled'). Note that you must be a super user to do this
2. Login
3. Generate a key and note the new button "Text key to patient"
4. Fill out your mobile phone number and send the text
5. Ensure that you receive a text message
6. You should be taken to a success page with options to navigate to other pages, ensure that the navigation works


covid_key/templates/covid_key/otk_sms_sent.html
```diff
@@ -0,0 +1,41 @@
...
+    <strong>Did {{ phone_number }} get the key?</strong>
+    <div>
+      <input type="radio" id="start" name="otk-sent-next" value="{% url 'start' %}" checked>
+      <label for="start">Yes, generate another key</label><br>
+      <input type="radio" id="otk_sms" name="otk-sent-next" value="{% url 'otk_sms' %}">
+      <label for="otk_sms">No, try different phone number</label><br>
+      <input type="radio" id="key" name="otk-sent-next" value="{% url 'key' %}">
+      <label for="key">No, read key</label>
+    </div>
+  </div>
```
pcraig3, line 37: Can we make this a real form? We can submit it as a POST request to the same URL and then redirect to whatever the submitted `value` is. We try to avoid JS where possible.

You don't have to use `cds_form` if it's too much work. We can just use `<form><fieldset><radios ...><button></form>` since it's the only place we have radios.

pcraig3: Also we would want to translate the text as well at some point. :)

profiles/static/js/script.js
```diff
@@ -19,3 +19,10 @@ if (form) {
   });
 }
 
+var radios = document.getElementsByName('otk-sent-next');
+radios.forEach(function(radio) {
+    radio.addEventListener('click', function(e) {
+        let value = document.querySelector('input[name="otk-sent-next"]:checked').value;
+        document.getElementById('otk_sms_sent_next').href = radio.value;
+    });
+});
```
pcraig3, line 28: This is pretty clever actually, but (mentioned above), we try to use default form interactions instead of JS where possible. We can just rely on the POST request and remove this.

pcraig3 (changes requested):
Hello! Looks really good overall! 

I've left a bunch of comments here and there: just things that we will want when we eventually merge this into the main branch. 

Let me know if there's anything you want to discuss in particular. At CDS we've been pretty bad at formalizing conventions (eg, form handling, input elements) in a way that's clearly written down, so apologies for the "we tend to do it this way" kinds of comments.

covid_key/forms.py
```diff
@@ -0,0 +1,40 @@
...
+from datetime import datetime
+
+from phonenumber_field.formfields import PhoneNumberField
+
+from dependency_injector.wiring import inject, Provide
+from portal.containers import Container
+from portal.services import NotifyService
+
+
+class OtkSmsForm(forms.Form):
```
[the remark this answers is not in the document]
andrej-vasilj: Great tip!

covid_key/forms.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    A form to specify patient phone number for receiving one time key (otk) by SMS
+    """
+
+    phone_number = PhoneNumberField(
+        label="Enter patient's mobile phone number", max_length=100
+    )
+    phone_number2 = PhoneNumberField(
+        label="Confirm patient's mobile phone number", max_length=100
+    )
```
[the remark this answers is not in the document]
andrej-vasilj: Not sure I understand...you mean in stead of "patient's" it should be "patient`s"? Is there a rationale for that?

covid_key/forms.py
```diff
@@ -0,0 +1,40 @@
...
+        otk,
+        notify_service: NotifyService = Provide[Container.notify_service],
+    ):
+        notify_service.send_sms(
+            phone_number=str(self.cleaned_data.get("phone_number")),
+            template_id=settings.OTK_SMS_TEMPLATE_ID.get(language or "en"),
+            details={
+                "code": otk["code"],
+                "expiry": datetime.fromtimestamp(otk["expiry"]).strftime(
+                    "%b %d %Y %H:%M:%S"
```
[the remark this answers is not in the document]
andrej-vasilj: Do you think we could just standardize this? Maybe we have the string formatting string as a setting in the settings file that we can then use everywhere?

covid_key/templates/covid_key/otk_sms.html
```diff
@@ -0,0 +1,18 @@
+{% extends 'base.html' %}
+{% load static i18n %}
+{% get_current_language as LANGUAGE_CODE %}
+
+{% block title %}{% trans "Text key to patient" %}{% endblock %}
+
+{% block content %}
+  {% include  "includes/flash_messages.html" %}
+  <h1>{% trans "Text key to patient" %}</h1>
```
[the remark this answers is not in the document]
andrej-vasilj: Sure, great tip. I don't think I'll bother trying to abstract that at this point, I'll just paste it in.

covid_key/forms.py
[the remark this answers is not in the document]
pcraig3: Instead of `patient's`, it should say `patient’s`.

covid_key/templates/covid_key/otk_sms_sent.html
andrej-vasilj: Ya I was wondering how you wanted to tackle this. I agree that a form-based solution would make sense.

covid_key/urls.py
```diff
@@ -9,18 +9,34 @@
...
     path(
         "generate-key/",
         login_required(
             otp_required(
-                TemplateView.as_view(template_name="covid_key/generate_key.html")
+                otp_required(
+                    TemplateView.as_view(template_name="covid_key/generate_key.html")
+                )
             )
         ),
```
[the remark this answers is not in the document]
andrej-vasilj: yes you're right, nice catch.

covid_key/urls.py
```diff
@@ -9,18 +9,34 @@
...
         name="generate_key",
     ),
+    path(
+        "otk_sms/",
+        login_required(otp_required(views.OtkSmsView.as_view())),
+        name="otk_sms",
+    ),
+    re_path(
+        r"otk_sms_sent/(?P<phone_number>[+\d]{0,50})$",
```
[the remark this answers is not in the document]
andrej-vasilj: Sure thing.

covid_key/views.py
```diff
@@ -9,23 +9,51 @@
...
+class StartView(TemplateView):
+    template_name = "covid_key/start.html"
+
+    def get(self, request):
+        # clear any existing one time keys
+        self.request.session.pop("otk", None)
+        return super().get(request)
+
+
+class SessionTemplateView(TemplateView):
```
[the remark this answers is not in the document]
andrej-vasilj: But that wouldn't describe it's behaviour though. It's a parent view class that checks if we have an otk cached within the session, that's why I called it "SessionTemplateView". I ended up using it in a few places...

profiles/static/js/script.js
andrej-vasilj: Makes sense to me

profiles/static/scss/components/_otk.scss
```diff
@@ -0,0 +1,36 @@
+div.otk {
+    display: inline-block;
+    position: relative;
+    font-size: xx-large;
+    background-color: $color-grey-footer;
+    border-radius: 4px;
+    padding: 10px 100px;
+    margin: 20px;
```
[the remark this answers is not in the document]
andrej-vasilj: I figured things would change there. It was more of a quick approach since it's a spike and since I have no idea how you guys wanted to standardize that

covid_key/forms.py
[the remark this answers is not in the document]
pcraig3: I think let's let it live here until we have "final" content. Once we do, it might make sense. For example, we are using a time format on the 'one-time-key' page.

covid_key/views.py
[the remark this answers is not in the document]
pcraig3: Okay, fair enough.

profiles/static/scss/components/_otk.scss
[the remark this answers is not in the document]
pcraig3: All good! We pretty much are following this pattern here: https://design-system.service.gov.uk/components/radios/#stacked-radios. 

The links I sent you before are an implementation in JS, but should be fairly portable. But if you wanted the source for the design / the HTML we would use, its the CDS Design System.

covid_key/forms.py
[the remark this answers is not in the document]
pcraig3: It's a typography "good practice" (re: snobby) thing: curly quotes are what you would see in a book or most other professionally printed material. Straight quotes originated in typewriters to save space and are really useful for coding, but they are kind of a stylistic no-no for content.

To take an excerpt from [the Butterick's Typography page](https://typographyforlawyers.com/straight-and-curly-quotes.html):

> Straight quotes are a _typewriter habit_. In traditional printing, all quotation marks were curly. But typewriter character sets were limited by mechanical constraints and physical space. By replacing the curly opening and closing quotes with ambidextrous straight quotes, two slots became available for other characters.

> Word processors are not limited in this way. You can always get curly quotes. Compared to straight quotes, curly quotes are more legible on the page and match the other characters better. Therefore, straight quotes should never, ever appear in your documents.

Basically it comes down to it's the convention we follow.

covid_key/tests.py
```diff
@@ -134,3 +137,58 @@ def test_CSRF_page_for_bad_posts(self):
...
+        """
+        self.login()
+        self.client.post(
+            reverse("key")
+        )  # generate a key so it can be cached in session
+        response = self.client.post(
+            reverse("otk_sms"),
+            {"phone_number": "+12125552368", "phone_number2": "+12125552368"},
+        )
+        self.assertEqual(response.status_code, 302)
```
pcraig3, line 180: We should also do a test with 2 different phone numbers (ie, that it's rejected).

covid_key/urls.py
```diff
@@ -23,4 +21,14 @@
...
     ),
+    path(
+        "otk-sms/",
+        login_required(otp_required(views.OtkSmsView.as_view())),
+        name="otk_sms",
+    ),
+    re_path(
+        r"otk-sms-sent/(?P<phone_number>[+\d]{0,50})$",
+        login_required(otp_required(views.OtkSmsSentView.as_view())),
+        name="otk_sms_sent",
```
pcraig3, line 32: Is the phone number here used for anything? I am not sure if it is.

portal/templates/includes/radio_select.html
```diff
@@ -0,0 +1,15 @@
+{% with id=widget.attrs.id %}
+<div {% if id %} id="{{ id }}"{% endif %} class="multiple-choice multiple-choice--radios">
+  {% for group, options, index in widget.optgroups %}
+    {% if group %}
+      <div class="multiple-choice--group">{{ group }}</div>
+    {% endif %}
+    {% for option in options %}
+      <div class="multiple-choice__item">
+        <input type="radio" name="{{ option.name }}" value="{{ option.value|stringformat:'s' }}">
+        <label {% if option.attrs.id %} for="{{ option.attrs.id }}"{% endif %}>{{ option.label }}</label>
```
pcraig3, line 10: The radios themselves look pretty good, but there's one small usability error: the `for` attribute of `label` has to be the same as the `id` of the `input` for the label to be properly associated with the input. 

So basically add `option.attrs.id` as the `id` of the input.

portal/widgets.py
```diff
@@ -0,0 +1,5 @@
+from django import forms
+
+
+class CDSRadioWidget(forms.widgets.RadioSelect):
+    template_name = "includes/radio_select.html"
```
pcraig3, line 5: This is cool, I haven't seen this done before.

covid_key/forms.py
```diff
@@ -0,0 +1,58 @@
...
+    A form to specify patient phone number for receiving one time key (otk) by SMS
+    """
+
+    phone_number = PhoneNumberField(
+        label="Enter patient’s mobile phone number", max_length=100
+    )
+    phone_number2 = PhoneNumberField(
+        label="Confirm patient’s mobile phone number", max_length=100
+    )
+
```
pcraig3, line 25: I should have noticed this, but I didn't.

There isn't currently validation to check that both numbers match each other. I tried entering two badly formatted phone numbers, and I got the error I expected, but if you enter 2 different phone numbers, the form submits correctly and sends a message to the first number entered. 

We should return an error message if the two numbers are different and also write a unit test for this.

pcraig3 (changes requested):
Okay! Looked again. Made a few comments but they shouldn't take long to fix 👍

covid_key/forms.py
andrej-vasilj: Ya for sure, nice catch

covid_key/urls.py
andrej-vasilj: I thought I would need it on the final page to say like "this number XXXXXXX was sent a text" but I guess not so I removed it.

covid_key/tests.py
```diff
@@ -144,6 +144,11 @@ def setUp(self):
         super().setUp()
         container.notify_service.override(NotifyService())  # Prevent sending emails/SMS
 
+        # Ensure that manitoba has sms enabled (test user is located in manitoba)
+        manitoba = HealthcareProvince.objects.get(name="Manitoba")
+        manitoba.sms_enabled = True
+        manitoba.save()
```
pcraig3, line 150: can we add a test that going to the `/otk-sms` url for a different province will redirect?

covid_key/forms.py
```diff
@@ -47,12 +47,7 @@ def send_sms(
         notify_service.send_sms(
             phone_number=str(self.cleaned_data.get("phone_number")),
             template_id=settings.OTK_SMS_TEMPLATE_ID.get(language or "en"),
-            details={
-                "code": otk["code"],
-                "expiry": datetime.fromtimestamp(otk["expiry"]).strftime(
-                    "%B %d %Y, %I:%M %p"
-                ),
-            },
+            details={"code": otk["code"]},
```
pcraig3, line 50: Hmm, we should also remove the "FR" template then, since we only have one for this one (and remove "EN" as a suffix"). 🎈 sorry 🍰

pcraig3 (commented):
Just a few small changes to suggest, but it looks good otherwise 👍

covid_key/forms.py
andrej-vasilj: I feel like there still may be some back and forth on that. I would suggest to keep it for now just in case.

andrej-vasilj:
@pcraig3 I figured out why I need the phone number param in the otk_sms_sent url. It's because on that page we display the text "did {{ phone_number }} receive the text?"

pcraig3 (commented):
Basically, the functionality works just fine, but there content for the [final designs](https://www.figma.com/file/znPqLTQQDcUkaNjdDcShiz/Exposure-Notification---Health-Portal?node-id=6677%3A47081) is quite often different from what's in this PR.

A lot of that predates this PR, for example, many of the changes to the page where the code is displayed initially are new. 

We should definitely update the text within the OTK flow so that it looks like the designs. For the other changes, that could be a separate card or it could be just bundled along here as well.

pcraig3:
Okay, slapped a commit on the end updating some of the words (and adding one more check to the form because it would throw an error if there was no value).

I think we could still look at formatting the number so that it looks better (`(416) 667 5327` instead of `+14166675327`), but let's just ship this ⛵

pcraig3 (approved):

Issue #356, opened 2021-01-12
Content change to account invite expiry screen


YedidaZalik:
@pcraig3 please change this screen to match the pink replacement screens in figma

For English: https://www.figma.com/file/znPqLTQQDcUkaNjdDcShiz/Exposure-Notification---Health-Portal?node-id=6375%3A38691

For French:  https://www.figma.com/file/znPqLTQQDcUkaNjdDcShiz/Exposure-Notification---Health-Portal?node-id=6385%3A2

I'll change the pink figma to black when you're done.  

Merci beaucoup!

pcraig3:
closed since #358

PR #358, opened 2021-01-14
Update error messages for expired/invalid invitations

pcraig3:
# Summary | Résumé

Updates the content for the "expired invite" messages, and some of the flash messages. Resolves issue #356 

### Test instructions | Instructions pour tester la modification

The best thing to check is if you get the right content (in English and French). Basically if you go to `/invite/expired` you;ll see the page content (but not the flash message).

## Screenshots

| before | after |
|--------|-------|
| Screen Shot 2021-01-14 at 12 56 31 PM   |  Screen Shot 2021-01-14 at 10 17 55 AM    |


andrej-vasilj (approved):
LGTM

PR #359, opened 2021-01-19
Add NITHA as a new province

pcraig3:
# Summary | Résumé

It's a Saskatchewan-based indigenous health authority. Adding it via a migration because we don't allow adding/editing provinces via the Django admin. Updated the provinces so that the abbreviation can be 5 characters long.

Also I renamed the last migration we had to give it a more human-friendly name.

### Test instructions | Instructions pour tester la modification

Log into the Django admin backend and check `Profiles` > `Provinces`. You should see NITHA in there as a new province.


bryan-robitaille (approved):
Checked PR deployment for newly added 'province'.  Checked renamed migration is properly referenced to.
Good to go!

pcraig3:
Merci!

PR #361, opened 2021-01-21
Added cron job for clearing audit logs

andrej-vasilj:
# Summary | Résumé

Clearing audit logs that are older than 180 days using a cron job and a custom Django management command.

---

This PR includes a modification to the Dockerfile to include supervisord in order to be able to run multiple processes within the same container. This allows us to run the cron daemon in addition to Django. A cron job was added that will run a new custom Django management command to clear all audit logs older than 180 days.

There are two possibilities to test this work:

1. Wait for it to be released on staging and ensure that all logs older than 6 months are automatically removed

or

2. Run this pull request locally which will allow you to change the parameters to make testing easier.
    i. In the file profiles/management/commands/clearauditlogs.py go to line 11 and change the time delta to something much smaller (like 5 days)
   ii. In the cronjobs.txt file change the job to run every minute (* * * * *) instead of (0 */2 * * *)
  iii. Build and run the system using docker-compose
  iv: Ensure that audit logs are deleted accordingly from your local dev database.


Dockerfile
```diff
@@ -19,7 +21,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 WORKDIR /code
 
 # Install pipenv
-RUN pip install 'pipenv==2018.11.26' uwsgi
+RUN pip install 'pipenv==2020.11.15' uwsgi
```
pcraig3, line 24: Seems to still work eh? I initially pinned the version for some or other reason, but I can't remember why anymore.

cronjobs.txt
```diff
@@ -0,0 +1,2 @@
+# Run every 2 hours
+0 */2 * * * /usr/local/bin/python3 /code/manage.py clearauditlogs > /proc/1/fd/1 2>/proc/1/fd/2
```
pcraig3, line 2: Maybe we should do once a day or once every 12 hours? We really don't need to run this often.

profiles/management/commands/clearauditlogs.py
```diff
@@ -0,0 +1,20 @@
...
+        delta = timezone.now() - timedelta(days=180)
+        AccessAttempt.objects.filter(attempt_time__lt=delta).delete()
+        AccessLog.objects.filter(attempt_time__lt=delta).delete()
+        RequestEvent.objects.filter(datetime__lt=delta).exclude(
+            url__in=["/en/key/", "/fr/key/"]
+        ).delete()
+        LoginEvent.objects.filter(datetime__lt=delta).delete()
+        CRUDEvent.objects.filter(datetime__lt=delta).exclude(
+            object_repr__startswith="COVIDKey object"
+        ).delete()
```
pcraig3, line 20: Nice that we can get them all in one go.

pcraig3 (commented):
I tested it and it all works just fine, so that's awesome. 

I had initially imagined it would be closer to Django itself. Inside of Axes, they will prune access attempts based on new logins: https://github.com/jazzband/django-axes/blob/0256c94720b7c9ab886d0957b4b79303f4a3068e/axes/attempts.py#L65

So, for example, we could add some kind of "delete" event to our custom audit backend here: https://github.com/cds-snc/covid-alert-portal/blob/500d49cad32df93973bf321ff6bb055a9ce923af/portal/audit_backends.py

eg
```
class LoggerBackend:
    def request(self, request_info):
        RequestEvent.objects.filter('older than 6 months')delete()
        # No need to log into the console for request, the WAF takes care of this
        return RequestEvent.objects.create(**request_info)
```

The pros of what you've done here is that we can see them all in one place.
The cons are having different functionality in different places.

Personally, I think the way you've done it works totally fine, so not an issue for me. 

Next steps:
- [ ] Dave should have a quick look
- [ ] Mark from TBS responds to my email

PR #363, opened 2021-01-29
Build "register" page, confirmation page, and "registrant" model

pcraig3:
# Summary | Résumé

Added a new "register" app, a new "Registrant" model, with a start page and a confirmation page.

- add start page
- add confirmation page
- add models for
  - Registrant details
- add page-by-page flow for registering a person (email and name) 

page-by-page flow includes:

1. add email
2. add name

Things we are skipping for this story: 
- Email confirmation
- Adding a location

Also, added a feature flag so that we can ship this to prod without it being visible.

## How to test this feature

- If you build it locally, make sure the `QR_CODES` waffle switch is active. 
- Go to url "/register"
- Click start
- Put in an email address
- Put in a name
- Confirmed!

## gif

Screen Recording 2021-01-29 at 10 04 06 AM


pcraig3:
fixed the unused imports

register/tests.py
```diff
@@ -0,0 +1,63 @@
...
+class RegisterView(TestCase):
+    def setUp(self):
+        Switch.objects.create(name="QR_CODES", active=True)
+
+    def test_start_page(self):
+        response = self.client.get(reverse("register:start"))
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(
+            response,
+            "<h1>Create a COVID Alert QR code for your venue</h1>",
```
andrej-vasilj, line 19: Just for awareness I would point out that this testing style isn't ideal as it creates an unnecessary coupling between the test and the code being tested. As in if we change the template content the test breaks. We can leave it for now, but at some point we should switch to something a bit more robust like 'assertTemplateUsed' where we just specify what template we expect to be rendered.

andrej-vasilj (approved):
Looks solid! Great initial effort! Thanks for the big push on your last days Paul!

dsamojlenko (approved):
Looks good!

register/tests.py
pcraig3: It's a good point, and we've definitely seen it before. Good suggestion!

PR #364, opened 2021-01-29
Update Github Actions to use newer versions

pcraig3:
In particular, we're not allowed to use short SHA's anymore to reference particular actions, so this change is necessary before Feb 15.

Shouldn't affect anything: deploys should continue to work as normal.


dsamojlenko:
Closing in favour of #368

PR #367, opened 2021-01-29
Add Location model, build Location forms, summary screen, save

dsamojlenko:
# Summary | Résumé

Add a location model to the new "register" app.

Fields for the location model can be seen on [the "summary" page](https://www.figma.com/proto/0CvVmmVb0EhnCmChzel70P/QR-Code-Registration-Website?node-id=7%3A520&scaling=min-zoom). 
Note that contact info is part of the location.

Build out Location flow:
- Add location category
- Add location name
- Add location address
- Add location contact info 
- Summary page for location and for a registrant

The Summary screen allows you to jump back and edit previously entered info. Upon submission at the Summary screen, the data will be saved to the Location table, and you will be directed to the Confirmation screen.

## How to test

- If you build it locally, make sure the QR_CODES waffle switch is active.
- Go to url "/register"
- Click start
- Fill out Registrant info (2 steps)
- Fill out Location info (4 steps)
- Summary
- Confirmed!

## Not implemented
- Email verification
- Back button
- Post code lookup


register/views.py
```diff
@@ -26,22 +29,82 @@ def form_valid(self, form):
...
+
+    def get_step_url(self, step):
+        return reverse(
+            self.url_name, kwargs={"pk": self.kwargs.get("pk"), "step": step}
+        )
+
+    def get_context_data(self, form, **kwargs):
+        context = super(LocationWizard, self).get_context_data(form=form, **kwargs)
+        registrant = Registrant.objects.get(id=self.kwargs.get("pk"))
+        context["form_data"] = self.get_all_cleaned_data
```
andrej-vasilj, line 76: don't you need to call the function here instead of referencing it? i.e. self.get_all_cleaned_data()?

andrej-vasilj (commented):
Looks great Dave! Awesome PR!

Just a few small comments:
1. If I go back in my browser after the first step, it seems that it doesn't populate my initial email address...
2. It seems that it doesn't validate the phone number syntax. I'm sure we could copy/paste that from another form where we input phone numbers.

register/views.py
dsamojlenko: Added the (), seems to work the same 🤷

andrej-vasilj (approved):

PR #368, opened 2021-02-01
fix: update amazon-ecs-deploy-task-definition

CalvinRodo:
Pin to long SHA's instead of short SHA's and Git Tags.

Pinning to a full length commit SHA is recommended in [Github's Security
Hardening Guide for Actions](https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions)

This should have no impact on the functionality of the Portal


andrej-vasilj (approved):
Looks fine to me...

dsamojlenko (approved):

PR #370, opened 2021-02-08
Refactor to support QR Codes as standalone app

andrej-vasilj:
# Summary | Résumé

Refactored the Portal Django project so that it can switch between serving the OTK Portal application and the new QR Code registration application.

Added AWS infrastructure so that the QR Code registration application can be served from an independent ECS service.

This deployment can be tested:

1. locally by setting the APP_SWITCH environment variable to either 'PORTAL'|'QRCODE'|'UNIT'. Based on this env variable the entry point will be either the otk portal or the new registration app
2.  using the new dev environment links:
portal.dev.covid-hcportal.cdssandbox.xyz
register.dev.covid-hcportal.cdssandbox.xyz

otherwise, it will only be testable on staging once merged


portal/settings.py
```diff
@@ -138,14 +138,27 @@
...
+    TEMPLATES[0]["OPTIONS"]["context_processors"] += [
+        "announcements.context_processors.announcement",
+        "profiles.context_processors.logout_messages",
+        "profiles.context_processors.general_settings",
+    ]
+elif APP_SWITCH == "QRCODE" or APP_SWITCH == "UNIT":
+    TEMPLATES[0]["OPTIONS"]["context_processors"] += [
+        "register.context_processors.general_settings",
+    ]
+
```
dsamojlenko, line 161: This conditional structure is a bit odd. If env var `APP_SWITCH=UNIT`, the second block will never run. Is the idea here that in the case of UNIT tests, all context_processors should be loaded? If so, then I don't think that's what's happening here.

dsamojlenko (approved):
Looks good 🤞

PR #371, opened 2021-02-08
Enable link from Confirmation screen to generate a new poster

dsamojlenko:
# Summary | Résumé

Enables the link on the Confirmation screen to "Generate a different poster" (previously was just linking back to the Start.

Also includes a minor rendering change for the Confirmation screen - previously it was rendering the confirmation from the POST route after submitting from the Summary. This adds a redirect to a confirmation-specific GET route.

## To test
- Start a new registration
- On the final Confirmation screen, click "Generate a different poster" and you will be taken to the Category step (instead of Start)

### DIFF
--- a/register/views.py
+++ b/register/views.py
@@ -1,7 +1,6 @@
 from django.views.generic import TemplateView, FormView
 from django.views.generic.edit import UpdateView
 from django.urls import reverse_lazy, reverse
-from django.shortcuts import render
 
 from waffle.mixins import WaffleSwitchMixin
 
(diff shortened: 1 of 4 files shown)


register/views.py
```diff
@@ -80,9 +80,8 @@ def get_context_data(self, form, **kwargs):
     def done(self, form_list, form_dict, **kwargs):
         forms = [form.cleaned_data for form in form_list]
         location = dict(ChainMap(*forms))
-        registrant = Registrant.objects.get(id=self.kwargs.get("pk"))
 
-        Location.objects.get_or_create(
+        Location.objects.create(
```
andrej-vasilj, line 84: How come we're switching from get_or_create to create? Don't we want to just fetch the same location if it already exists? I'm also thinking that maybe we should add some uniqueness constraints on the model itself to prevent duplicates, no?

andrej-vasilj (approved):

register/views.py
dsamojlenko: Yeah, I think that's where we want to go, but that's not what this copy-pasted code was achieving... If we're going to detect/resolve duplicate addresses, it'll include changing the UI to assist in that. This just was just causing problems in testing.

PR #373, opened 2021-02-08
Add unique short_code to Location model

dsamojlenko:
# Summary | Résumé

This PR adds a short_code column to the Location model, and then overrides the model's save() method to generate a unique short_code for the Location on object create.

The goal here is to have a shorter id as a reference to the Location record, rather than the 128-bit UUID that is the primary key.

The main reason for having a shortened key is for the sake of having a smaller payload for the QR code, making it less complex, and therefore easier to scan.


register/models.py
```diff
@@ -29,6 +33,16 @@ class Location(models.Model):
     contact_email = models.EmailField(verbose_name=_("Email address"), max_length=255)
     contact_phone = PhoneNumberField(blank=True)
 
+    # Override save method to generate a unique short code for poster
+    def save(self, *args, **kwargs):
```
andrej-vasilj, line 37: I like this! Slick placing of the logic for generating unique codes

andrej-vasilj (approved):
Looks perfect, nice PR!

PR #380, opened 2021-02-15
Add registrant email verification

dsamojlenko:
# Summary | Résumé

Adds the Registrant email verification step. The flow is as follows:

- User enters email address
- System creates email_confirmation record
- System sends confirmation email
- User clicks link in email
- System ensures confirmation link is valid (expires after 24hrs)
- System creates or gets existing Registrant
- User is redirected to the Name step (pre-populated if registrant already exists)
- System deletes the email_confirmation record
- User continues with Location registration

To test:
- You will need the template_ids for the new Notify Confirmation Email templates (see .env.example, Notify)


register/templates/register/error.html
```diff
@@ -0,0 +1,11 @@
+
+{% extends 'base.html' %}
+{% load i18n %}
+
+{% block title %}
+  Not Logged in
```
andrej-vasilj, line 6: Not sure if this title property is really relevant for anything, but it should probably have a similar meaning/context as the message...maybe something like "Link expired".

register/templates/register/registrant_email_submitted.html
```diff
@@ -0,0 +1,20 @@
...
+{% block content %}
+  {% include  "includes/error_list.html" %}
+
+  <h1>{% trans "Confirm your email address" %}</h1>
+  
+  {% blocktrans with email=submitted_email %}
+    <p>We sent a link to {{ email }}.</p>
+  {% endblocktrans %}
+
+  <p>{% trans "If you did not get the email, check your junk or spam folder." %}
```
andrej-vasilj, line 18: Do we have a "send again" functionality planned for this?

register/views.py
```diff
@@ -17,17 +22,67 @@ class RegistrantEmailView(FormView):
...
 
-        self._object = obj
-        self.request.session["registrant_email"] = email
+        form.send_mail(self.request.LANGUAGE_CODE, email, url)
 
         return super().form_valid(form)
 
     def get_success_url(self):
-        return reverse_lazy("register:registrant_name", kwargs={"pk": self._object.pk})
+        self.request.session["submitted_email"] = self.submitted_email
```
andrej-vasilj, line 42: I'm wondering if you could just move this line to the function above this. Then you could remove this function all together and use the "success_url" class property instead. It would be a bit more concise. Generally with Django I would only override a class property with its functional form when I need to add extra logic.

register/views.py
```diff
@@ -17,17 +22,67 @@ class RegistrantEmailView(FormView):
...
+        confirm.delete()
+
+        return redirect(
+            reverse_lazy("register:registrant_name", kwargs={"pk": registrant.pk})
+        )
+    except (EmailConfirmation.DoesNotExist):
+        return redirect(reverse_lazy("register:confirm_email_error"))
+
+
+class RegistrantEmailConfirmError(TemplateView):
```
andrej-vasilj, line 84: There's no need for this class. You can just use a template view directly in the url definition.

register/views.py
```diff
@@ -17,17 +22,67 @@ class RegistrantEmailView(FormView):
...
+        del self.request.session["submitted_email"]
+        return context
+
+
+def confirm_email(request, pk):
+    try:
+        # Confirmation emails are only valid for 24hrs
+        time_threshold = datetime.now() - timedelta(hours=24)
+
+        confirm = EmailConfirmation.objects.get(id=pk, created__gt=time_threshold)
```
andrej-vasilj, line 68: While this will definitely work, I think it would be better to get the instance and then explicitly perform the datetime comparison. By using the ORM to perform this comparison we wouldn't know if the exception is due to the date being expired or if for some strange reason the PK was missing from the DB. In this latter case we could still show the user the same error page or maybe even just crash with an error 500. In either case I think it would be nice to be able to log that event if it happened. Like if for example somebody was just trying random PKs as some form of attack? Just thinking out loud here...

register/views.py
```diff
@@ -17,17 +22,67 @@ class RegistrantEmailView(FormView):
...
+        request.session["registrant_email"] = confirm.email
+
+        # Create the Registrant
+        registrant, created = Registrant.objects.get_or_create(email=confirm.email)
```
andrej-vasilj, line 72: I think we could use the "created" flag here to our advantage and save the registrant's name to the session as well if it already exists. This way we could pre-populate it into the template to save the user a step if they are creating multiple qr codes.

portal/.env.example
```diff
@@ -19,6 +19,9 @@ API_ENDPOINT = https://example.com/new-key-claim
 # PASSWORD_RESET_EMAIL_TEMPLATE_ID_EN =
 # CONFIRMATION_EMAIL_TEMPLATE_ID_EN =
 # CONFIRMATION_EMAIL_TEMPLATE_ID_FR =
+# REGISTER_EMAIL_CONFIRMATION_ID_EN =
```
andrej-vasilj, line 22: Currently if we add a new notify template we have to add it to this env example file and the settings which you have already, and the readme plus like 5 or 6 different infrastructure related files. On top of this we need to remember to add the secret to Github directly (in staging and production) and manually to Heroku. If I remember correctly Paul made a document for this process a few days before you joined. I have no idea where this document is living but maybe Brian would know.

andrej-vasilj (changes requested):
Nice work overall Dave! I have a few fairly minor comments. Mostly we need to ensure that all of the infrastructure related files are updated to get the new notify template to work.

register/views.py
dsamojlenko: Refactored a bit to account for logging the different possible errors (expired vs invalid). Still rendering the same error page.

register/views.py
dsamojlenko: So the Name field already does pre-populate for returning users - not from the session though, from the model- this was leftover from the previous iteration.

register/views.py
dsamojlenko: sweet. made the same change to the start page as it was setup the same way.

register/templates/register/error.html
dsamojlenko: We don't have content for this page yet. I've updated it for context to "Invalid or expired link" but that should be replaced by final content later.

register/templates/register/registrant_email_submitted.html
dsamojlenko: ah yes, missed that button.

portal/.env.example
dsamojlenko: I will hunt that doc down

andrej-vasilj (approved):
Looks good Dave, thanks for the changes! Are the tests coming afterwards in another PR?

dsamojlenko:
Merging. Tests will come in another PR.

PR #381, opened 2021-02-19
Alerting flow initial pass

andrej-vasilj:
# Summary | Résumé

Added a new Django app for the outbreak alerting flow

Login as either a superadmin or a user that has the new permission "can_send_alerts" and go through the flow from the new "Public exposures" nav tab.

After creating a new exposure alert, check that you can see it in the exposure history page.


portal/templates/includes/nav_main.html
```diff
@@ -11,6 +11,9 @@
         <li><a {% if request.resolver_match.url_name == 'profiles' %}class="active"{% endif %} href="{% url 'profiles' %}">{% trans "Manage team" %}</a></li>
       {% endif %}
       <li><a {% if request.resolver_match.url_name == 'support' %}class="active"{% endif %} href="{% url 'support' %}">{% trans "Support" %}</a></li>
+      {% if perms.profiles.can_send_alert %}
+        <li><a {% if request.resolver_match.url_name == 'public_exposures/start' %}class="active"{% endif %} href="{% url 'exposure_notifications:start' %}">{% trans "Public exposures" %}</a></li>
```
dsamojlenko, line 15: This active url match is incorrect (highlight doesn't appear under "Public exposures" when in that section)

dsamojlenko:
I wonder about naming the new app "exposure_notifications" given that has a very specific meaning in the COVID Alert ecosystem. Even though this is not visible to the end user, using consistent terminology is important.

outbreaks or outbreak_notifications maybe?

exposure_notifications/templates/confirm.html
```diff
@@ -0,0 +1,58 @@
...
+{% load i18n %}
+
+{% block title %}
+    {% trans "Confirm notification" %}
+{% endblock %}
+
+{% block content %}
+    <div class="content-back-link">
+        <a href="{% url 'exposure_notifications:severity' %}" class="chevron-left">{% trans "Back" %}</a>
+    </div>
```
dsamojlenko, line 11: I ended up making a back-link include in my PR ... I am going to do this instead.

exposure_notifications/templates/confirm.html
```diff
@@ -0,0 +1,58 @@
...
+        <tr>
+            <td>{% trans "Date(s)" %}</td>
+            <td>{{ alert_datetime }}</td>
+            <td><a href="{% url 'exposure_notifications:datetime' %}"><span>{% trans "change" %}</span></a></td>
+        </tr>
+        <tr>
+            <td>{% trans "Severity" %}</td>
+            <td>{{ alert_level }}</td>
+            <td><a href="{% url 'exposure_notifications:severity' %}"><span>{% trans "change" %}</span></a></td>
+        </tr>
```
dsamojlenko, line 50: "Change" should be capitalized.

Also, see example here of how we can give better link context to non-visual users: https://github.com/cds-snc/covid-alert-portal/blob/main/register/templates/register/summary.html#L56

ie:

```
<a href="">Change <span class="visually-hidden"> thing to be changed</span></a>
```

exposure_notifications/templates/details.html
```diff
@@ -0,0 +1,67 @@
...
+    </div>
+
+  {% if location %}
+    <h1>{% trans "Details of public exposure" %}</h1>
+
+    <table>
+      <caption class="visually-hidden">{% trans "Exposure details" %}</caption>
+      <tbody>
+        <tr>
+            <td>{% trans "Name" %}</td>
```
dsamojlenko, line 20: these row headings should be marked up as headings with scope for a11y:

`<th scope="row">{% trans "Name" %}</td>

exposure_notifications/templates/details.html
```diff
@@ -0,0 +1,67 @@
...
+            <td>{{ location.name }}</td>
+            <td></td>
+        </tr>
+        <tr>
+            <td>{% trans "Address" %}</td>
+            <td>
+                <span> {{ location.address }} </span>
+                <span> {{ location.city }}, {{ location.province }} </span>
+                <span> {{ location.postal_code }} </span>
```
dsamojlenko, line 29: Maybe consider wrapping this in an ```<address>``` element https://developer.mozilla.org/en-US/docs/Web/HTML/Element/address

exposure_notifications/templates/details.html
```diff
@@ -0,0 +1,67 @@
...
+            <td></td>
+        </tr>
+        <tr>
+            <td>{% trans "Date of public exposure" %}</td>
+            <td>{{ notification.start_date|date:'d/m/Y' }}</td>
+            <td></td>
+        </tr>
+        <tr>
+            <td>{% trans "Date of notification" %}</td>
+            <td>{{ notification.created_date|date:'d/m/Y' }}</td>
```
dsamojlenko, line 52: As mentioned in the review earlier, strongly suggest styling date as yyyy-mm-dd

dsamojlenko:
Noting that on the Public exposures Search form, if I ignore the browser "required" attribute and submit the search, we get an error. Probably needs some form validation. 
image

exposure_notifications/templates/datetime.html
```diff
@@ -0,0 +1,56 @@
...
+          {% endif %}
+          {{ form.month }}
+        </div>
+        <div class="date_entry">
+          {{ form.year.label_tag }}
+          {% if form.year.errors %}
+            <div id="error--{{ form.year.html_name }}" class="validation-error">{{ form.year.errors|first }}</div>
+          {% endif %}
+          {{ form.year }}
+        </div>
```
dsamojlenko, line 42: Formatting here gets a little screwy on validation errors. Probably a case where you can roll all the validation errors into one

image

exposure_notifications/templates/search.html
```diff
@@ -0,0 +1,90 @@
...
+          <tr>
+            <td>{{ location.name }}</td>
+            <td>
+              <span> {{ location.address }} </span>
+              <span> {{ location.city }}, {{ location.province }} </span>
+              <span> {{ location.postal_code }} </span>
+            </td>
+            <td>
+              <a href="{% url 'exposure_notifications:profile' location.id %}">
+                <span aria-hidden="true">{% trans "Send alert" %}</span>
```
dsamojlenko, line 53: This is another one of those situations where we should add some visually-hidden text to help with accessibility.

And I'm pretty sure you didn't mean to make that "Send alert" link invisible to screen readers?

exposure_notifications/forms.py
```diff
@@ -0,0 +1,39 @@
...
+class SeverityForm(forms.Form):
+    alert_level = forms.ChoiceField(
+        label=_("Select level of public exposure"),
+        choices=[
+            ("1", _("Somebody sneezed...")),
+            ("2", _("Sirens and lights flashing everywhere!")),
+            ("3", _("Zombie apocalypse!!!!")),
+        ],
+        widget=CDSRadioWidget(attrs={"class": "multichoice-radio"}),
+    )
```
dsamojlenko, line 39: I think you want the label here to be blank. What ends up happening is this label is rendered and associated with the first radio element, which in turn also has a label attached to it. We get an Axe violation because an input can't have multiple label elements.

Typically, I think we've been leaving this label element empty and using the page title as a descriptive title for the radio list.

image

image

exposure_notifications/templates/history.html
```diff
@@ -0,0 +1,81 @@
...
+            <td>
+              <span> {{ notification.location.address }} </span>
+              <span> {{ notification.location.city }}, {{ notification.location.province }} </span>
+              <span> {{ notification.location.postal_code }} </span>
+            </td>
+            <td>{{ notification.start_date|date:'d/m/Y' }}</td>
+            <td>
+              <a href="{% url 'exposure_notifications:details' notification.id %}">
+                <span aria-hidden="true">{% trans "Details" %}</span>
+              </a>
```
dsamojlenko, line 54: Another case where we can provide some visually-hidden text to help non-visual users, and that Details text should not be hidden with aria-hidden.

exposure_notifications/templates/profile.html
```diff
@@ -0,0 +1,61 @@
...
+    </div>
+
+  {% if location %}
+    <h1>{% trans "Place profile" %}</h1>
+
+    <table>
+      <caption class="visually-hidden">{% trans "Location profile" %}</caption>
+      <tbody>
+        <tr>
+            <td>{% trans "Name" %}</td>
```
dsamojlenko, line 20: more heading rows in this table, these should be marked up as `<th scope="row">...`

portal/templates/includes/nav_main.html
andrej-vasilj: Nice catch!

exposure_notifications/templates/details.html
andrej-vasilj: I changed this all around

andrej-vasilj:
> Noting that on the Public exposures Search form, if I ignore the browser "required" attribute and submit the search, we get an error. Probably needs some form validation.
> image

I already fixed this in my current work that will be an upcoming PR so I'd prefer not to try to extract it out and add it here, but let's keep a note of this since it's an important piece for sure.

exposure_notifications/templates/datetime.html
andrej-vasilj: This I will also fix in my upcoming PR since I refactored a lot of stuff regarding this form and noticed this problem in the process

exposure_notifications/forms.py
andrej-vasilj: Great catch!

andrej-vasilj:
Thanks for all of the great accessibility feedback @dsamojlenko, I definitely learned a few new things!

outbreaks/templates/confirm.html
```diff
@@ -0,0 +1,58 @@
...
+                </address>
+            </td>
+            <td>
+                <a href="{{ map_link }}" target="_blank">{% trans "Show on map" %}</a>
+            </td>
+        </tr>
+        <tr>
+            <th scope="row">{% trans "Type" %}</th>
+            <td>{{ location.category }}</td>
+            <td><a href="{% url 'outbreaks:search' %}">{% trans "Change" %}<span class="visually-hidden">{% trans "location" %}</span></a></td>
```
dsamojlenko, line 39: the visuall-hidden text here should be "category"

dsamojlenko (approved):

outbreaks/templates/confirm.html
andrej-vasilj: actually I don't think so...We can't change the actual location category from the alert portal, we can only change the location

PR #382, opened 2021-02-19
Update registration flow content

dsamojlenko:
# Summary | Résumé

Updates Location registration flow to latest content and design. Includes:
- New base.html template for Register site
- New custom scss styles for Register site
- Updates to content and flow
- Added contact_name property to Location
- Removed registrant_id from URL schemes (storing in session)
- New footer.html include for Register site
- Back link include for Register site
- Custom dict_key filter for parsing dictionaries by dynamic key in templates
- Customized radio_select widget to accept optional help_text for radio items


portal/templates/includes/radio_select.html
```diff
@@ -7,9 +7,14 @@
     {% for option in options %}
       <div class="multiple-choice__item">
         <input {% if option.attrs.id %} id="{{ option.attrs.id }}"{% endif %} type="radio" name="{{ option.name }}" value="{{ option.value|stringformat:'s' }}" {% if option.attrs.checked %} checked{% endif %}>
-        <label {% if option.attrs.id %} for="{{ option.attrs.id }}"{% endif %}>{{ option.label }}</label>
+        <label {% if option.attrs.id %} for="{{ option.attrs.id }}"{% endif %}>
+          {{ option.label }}
+          {% if option.attrs.help_text %}
```
andrej-vasilj, line 12: Great solution! I was thinking along the same lines for this issue

register/templates/register/confirmation.html
```diff
@@ -1,34 +1,65 @@
-{% extends 'base.html' %}
+{% extends 'register/base.html' %}
 {% load i18n %}
 
 {% block title %}
-  {% trans "Your request for a poster has been submitted" %}
+  {% trans "We've got the information for your poster" %}
```
andrej-vasilj, line 5: iirc Paul told me that for whatever reason we should use a diff character for the apostrophes. He recommended ’ instead of '. Not sure if you know anything about that or want to continue that...

register/templates/register/confirmation.html
```diff
@@ -1,34 +1,65 @@
...
+  <h2>{% trans "Next steps" %}</h2>
 
-  <div class="info-message">
-    {% trans "If your business organization or event operates across multiple branches, you will need to create a poster for each unique location." %}
+  <div class="next-steps">
+    <div>
+      <p>{% trans "<a href=''>Contact us</a> if you do not get the email within X hours or if you have questions." %}</p>
+      <p>{% trans "Remember to make more posters if your service or event has other addresses. You can come back later if you do not have time now." %}</p>
+    </div>
+    <svg width="104" height="132" viewBox="0 0 104 132" fill="none" xmlns="http://www.w3.org/2000/svg">
```
andrej-vasilj, line 29: Can we maybe serparate out the svg images into their own asset files just for clarity? It's a bit odd to see them directly embedded into the template

register/tests.py
```diff
@@ -15,33 +15,19 @@ def setUp(self):
     def test_start_page(self):
```
andrej-vasilj, line 15: You may want to test that the correct template is being rendered for these.

register/tests.py
```diff
@@ -15,33 +15,19 @@ def setUp(self):
...
         response = self.client.get(reverse("register:registrant_email"))
         self.assertEqual(response.status_code, 200)
-        self.assertContains(
-            response,
-            "<h1>Where should we send your poster?</h1>",
-            html=True,
-        )
 
     def test_name_page(self):
         r = Registrant.objects.create(email="test@test.com")
```
andrej-vasilj, line 24: I think the negative tests are useful as well. Like what should happen if we don't have the registrant id stored in the session.

register/views.py
```diff
@@ -109,13 +116,11 @@ def get_template_names(self):
         return [TEMPLATES[self.steps.current]]
```
andrej-vasilj, line 116: For this view and beyond we need to have some sort of safeguard that prevents people from accessing these links directly without having confirmed an email. Currently it will just crash with a KeyError when it tries to access the registrant_id from the session which doesn't exist.

andrej-vasilj (changes requested):
Great PR Dave! Thanks a lot for refactoring the base/footer templates for the register app! I have a few small comments that you can decide whether to do or not and one more crucial one at the end regarding the view access.

register/templates/register/confirmation.html
dsamojlenko: might as well, I usually just go with whatever I copy/paste out of designs... maybe I typed this one.

register/templates/register/confirmation.html
dsamojlenko: ... better for caching too. think I put them there for quickness/convenience, but forgot to move them out

register/tests.py
dsamojlenko: beauty

register/views.py
dsamojlenko: yeah. would a middleware approach be appropriate?

dsamojlenko: took a different approach

dsamojlenko:
Waiting on the Django patch in PR #381 to fix the Snyk error

register/tests.py
```diff
@@ -4,73 +4,155 @@
...
 from .utils import generate_random_key
+from django.contrib.messages import get_messages
+from portal.services import NotifyService
+from portal import container
 
 
 class RegisterView(TestCase):
     def setUp(self):
-        pass
+        Switch.objects.create(name="QR_CODES", active=True)
```
andrej-vasilj, line 16: Do we still need this switch for something?

register/views.py
```diff
@@ -132,18 +158,27 @@ def done(self, form_list, form_dict, **kwargs):
...
-        return HttpResponseRedirect(
-            reverse("register:confirmation", kwargs={"pk": self.kwargs.get("pk")})
-        )
+        return HttpResponseRedirect(reverse("register:confirmation"))
 
 
 class RegisterConfirmationPageView(TemplateView):
     template_name = "register/confirmation.html"
 
+    def dispatch(self, request, *args, **kwargs):
```
andrej-vasilj, line 172: Great that we have this now! I see that this same function is repeating a few times. It's not a big deal, so feel free to merge as-is, but it would be nice to have that in a single place. Maybe just subclass the TemplateView, add that dispatch method and inherit from it in all of your views?

register/tests.py
```diff
@@ -4,73 +4,155 @@
...
         session.save()
 
-        r = Registrant.objects.create(email=email)
+        response = self.client.get(reverse("register:confirmation"))
+        self.assertEqual(response.status_code, 200)
 
-        response = self.client.get(
-            reverse("register:confirmation", kwargs={"pk": r.pk})
+
+class RegisterEmailConfirmation(TestCase):
```
andrej-vasilj, line 51: That's a lot of new unit tests! Looking good!

andrej-vasilj (approved):
Looks great Dave, nice PR!. I have a few minor questions/comments, but feel free to merge as-is if you want.

register/tests.py
dsamojlenko: Good point, no we don't ... removed other references as well

register/views.py
dsamojlenko: yeah I was trying to figure out the "right way" to re-use, but since it was just a few places figured ok to roll for now, we can refactor later.

PR #384, opened 2021-02-23
French registration flow

ghost:
Added French translations to QR code registration flow pages.

Some of the translations weren't available in Figma and there seems to be some issue in translating data from *.py files. Looking into it at the moment. Most of the translations through html pages are complete.


register/templates/register/includes/back.html
```diff
@@ -1,7 +1,9 @@
+{% load i18n %}
+
 <div class="back-button"> 
     <svg width="10" height="17" viewBox="0 0 10 17" fill="none" xmlns="http://www.w3.org/2000/svg">
         <path d="M9 1.5L2 8.49999L8.99999 15.5" stroke="#2B4380" stroke-width="2"/>
     </svg>
 
-    <a href="{{ link }}">Back</a>
+    <a href="{{ link }}">{% trans "Back" %}</a>
```
dsamojlenko, line 8: Good catch!

dsamojlenko:
This is looking great - you caught a few spots where I missed using the translation function. There are a few "fuzzy" matches we can review, I'm assuming these are the missing translations - we'll talk about how to proceed.

dsamojlenko:
I changed the target on the PR so the diff looks better (it's now targeting the branch of PR #382)

dsamojlenko (approved):
Great work!

PR #385, opened 2021-02-25
last_modified_password_datetime_display

sakura-sketch:
Added "Last changed: <datetime>" to "Manage your account" with french translation


sakura-sketch:
image

The last updated date of the password is visible in the screenshot. The code gets the data from CRUDEvent.

After logging into your account, go to Manage your account -> Change password -> The last changed date gets updated.

Note: The "Last changed" is empty if the user has never updated the password

Pipfile
```diff
@@ -44,6 +44,8 @@ django-waffle = "*"
 django-solo = "*"
 dependency-injector = "*"
 django-formtools = "*"
+uwsgi = "*"
```
andrej-vasilj, line 47: I'm ok with adding uwsgi here, but we should then remove it from the dockerfile. I'm also curious to know why it was placed by itself in the dockerfile instead of here in the first place. Maybe @dsamojlenko you have an idea?

andrej-vasilj (dismissed):
Hey Nitya, nice work! Looks like it will work perfectly well, I just have some minor comments as I think that we could make this a bit more concise.

Pipfile
dsamojlenko: I am not sure ... what functionality does uwsgi provide?

profiles/views.py
```diff
@@ -385,6 +387,34 @@ def get_context_data(self, *args, **kwargs):
...
+            cur_lang = translation.get_language()
+
+            recent_date = (
+                CRUDEvent.objects.filter(user_id=self.request.user.id)
+                .filter(changed_fields__icontains="password")
+                .first()
+                .datetime
+            ).date()
+
+            get_day = str(recent_date.day)
```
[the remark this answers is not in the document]
sakura-sketch: I did use strftime but it was difficult to translate the month to French with the day and year in the string. Hence, this workaround. I am open to switching it back if I missed anything about translations

Pipfile
sakura-sketch: I always got a missing uwsgi dependency when I tried to run the server. I can remove it, I had to manually install it maybe that's why it showed up there.

dsamojlenko: Let's pull it out of this PR since it's unrelated, but we can talk about adding it to the Pipfile.

dsamojlenko:
I think in this situation, where I don't have any password changes in my history, we should probably not display the "Last changed" label.

image

profiles/views.py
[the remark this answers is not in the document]
dsamojlenko: I think Andrej is right, there's got to be a way to leverage strftime properly, I think you just need to look at the formatting options.

dsamojlenko (approved):

dsamojlenko (approved):

sakura-sketch (commented):
Performed the requested changes

PR #386, opened 2021-02-25
Integrate Canada Post Address Complete

dsamojlenko:
# Summary | Résumé

Integrate the Canada Post Address Complete API with the [GDS Accessible Autocomplete](https://github.com/alphagov/accessible-autocomplete) component, and use it in the Location Address form.

image


register/templates/register/location_address.html
```diff
@@ -21,3 +21,8 @@ <h1>{% trans "Address of service or event" %}</h1>
 
   {% include "includes/cds_form.html" with submit_text=_("Next") %}
 {% endblock %}
+
+{% block scripts %}
+  <script src="{% static 'register/js/accessible-autocomplete.min.js' %}"></script>
+  <script src="{% static 'register/js/address-autocomplete.js' %}"></script>
```
dsamojlenko, line 27: Maybe consider loading these through [jsDelivr](https://www.jsdelivr.com/)

register/static/register/js/address-autocomplete.js
```diff
@@ -0,0 +1,140 @@
+/**
+ * This is the Canada Post API key. Yes it's weird to see it here
```
andrej-vasilj, line 2: Thanks for the comment! That certainly does look weird otherwise :)

andrej-vasilj (approved):
Looks great to me Dave! I don't see any immediate issues with it.

PR #387, opened 2021-02-26
Datepicker

andrej-vasilj:
The date picker should now be working in the public exposures tab.


portal/templates/base.html
```diff
@@ -28,7 +28,12 @@
     <!-- Google Analytics-->
     {% include "includes/google_analytics.html" %}
 
+    <!-- Datepicker (set to last so that it takes precedence) -->
+    <script type="module" src="https://cdn.jsdelivr.net/npm/@duetds/date-picker@1.2.0/dist/duet/duet.esm.js"></script>
+    <script nomodule src="https://cdn.jsdelivr.net/npm/@duetds/date-picker@1.2.0/dist/duet/duet.js"></script>
```
dsamojlenko, line 33: Should these JS files be linked in the footer?

andrej-vasilj: I think it's fine here. I think I tried experimenting with some other locations for the imports but it didn't seem to be working.

dsamojlenko: I just mean because normally we put JS at the end of the file so it doesn't delay rendering of the HTML if there's a network problem. JS is render blocking.

portal/settings.py
```diff
@@ -408,7 +413,10 @@
 ]
 CSP_SCRIPT_SRC = [
     "'self'",
+    "'unsafe-eval'",
+    "'unsafe-inline'",
```
dsamojlenko, line 417: we should tighten this up. unsafe-eval and unsafe-inline open us up to cross-site-scripting attacks, and our Mozilla Observatory rating is taking a hit.

image

dsamojlenko: Just did a test locally, I think you just need to add `"blob:"` to the array instead of the two unsafe- entries.

portal/settings.py
```diff
@@ -399,7 +399,12 @@
     "covid-alert-portal.alpha.canada.ca",
     "portail-alerte-covid.alpha.canada.ca",
 ]
-CSP_STYLE_SRC = ["'self'", "fonts.googleapis.com"]
+CSP_STYLE_SRC = [
+    "'self'",
+    "'unsafe-inline'",
```
dsamojlenko, line 404: Should tighten this up too - not as dangerous as the script_src, but still an issue. I *think* you just need to specify the hashes for the inline styles, which Chrome conveniently suggests in the error messages:
image

So, in this case, instead of `"'unsafe-inline'"`, add:

```
    "'sha256-wj97CBnjU4ly4cqQnM0huquRQF943qGisXfpavz3YsU='",
    "'sha256-uwPeycnNJj5VRuojI2kTqwzlRkuYgdlJ3laHnGO8S64='",
```

portal/templates/base.html
dsamojlenko: I see in the docs they specifically recommend putting them in <head> ... if they won't work in the footer, maybe we can look at using the preload tag they recommend for optimizing for CDNs: https://github.com/duetds/date-picker#optimizing-cdn-performance

profiles/static/scss/components/forms/_input.scss
```diff
@@ -36,3 +36,17 @@ main form {
...
+    display: none;
+}
+
+.duet-date__dialog-content {
+    display: table;
+}
+
+.duet-date__close {
+    margin-right: 15px;
+}
```
dsamojlenko, line 52: Something in here is causing the Voice Over problem we saw earlier. If I remove them, obviously things look crazy, but I don't get stuck in that weird loop every time I open the date picker.

portal/templates/base.html
andrej-vasilj: I'll try moving it to the footer

portal/settings.py
andrej-vasilj: ok I'll try that out. I was kind of worried about those as well

portal/settings.py
andrej-vasilj: will do

outbreaks/templates/datetime.html
```diff
@@ -10,47 +10,68 @@
...
-          </button>
-        </div>
+          <button type="button" class="link datepicker_button" id="{{ fieldset.title }}">{% trans "select date" %}</button>
 
+          <duet-date-picker
+                  id="datepicker_{{ fieldset.title }}"
+                  value="{{ fieldset|fieldset_date }}"
+                  min="{{min_date}}"
+                  max="{{max_date}}"
+                  language="{{language}}"
```
dsamojlenko, line 53: I know we don't have all the translations for the page, but I see this language param being set here - but on the French page, the calendar is still in English 
image

dsamojlenko:
In the day-of-the-week heading, Sunday is missing, and also at the end of the week

image

dsamojlenko (approved):

PR #401, opened 2021-03-01
Render custom error templates from register app

dsamojlenko:
# Summary | Résumé

In a previous PR the Register site got its own Base template, however the default error pages for the site (500, 403, 404, etc) all extended the Portal base template, which has references to routes that do not exist in the Register app.

This PR introduces error pages specifically for the Register app, and tells the Portal to render the correct error template from the Register app when in Register mode.

### DIFF
--- a/portal/views.py
+++ b/portal/views.py
@@ -8,14 +8,20 @@ from urllib.parse import unquote, urlparse
 
 
 def permission_denied_view(request, exception=None):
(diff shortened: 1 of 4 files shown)


andrej-vasilj (approved):
I'm fine with this. Can't think of anything more elegant off the top of my head.

PR #403, opened 2021-03-03
Venue categories display summary

sakura-sketch:
Added changes to display location category label in the summary
Screen Shot 2021-03-02 at 8 32 47 PM


register/forms.py
```diff
@@ -37,17 +37,18 @@ class Meta:
...
 
 class LocationCategoryForm(HealthcareBaseForm, forms.Form):
+    location_choices = [
+        ("restaurant_bar_coffee", _("Restaurant, bar, coffee shop")),
+        ("fitness_recreation", _("Fitness and recreation")),
+        ("arts_entertainment", _("Arts and entertainment")),
+        ("grooming_wellness", _("Grooming and wellness")),
+        ("religious_space", _("Religious space")),
+        ("other", _("Something else")),
+    ]
```
dsamojlenko, line 47: Suggestion: move this out of the class, so it can be imported a little more cleanly.

register/views.py
```diff
@@ -12,6 +12,7 @@
 from datetime import datetime, timedelta
 import pytz
 from django.contrib import messages
+from .forms import LocationCategoryForm as choices
```
dsamojlenko, line 16: If you follow the suggestion to move the choices out of the form class, then this would be `from .forms import location_choices`

register/views.py
```diff
@@ -144,6 +145,11 @@ def get_context_data(self, form, **kwargs):
         registrant = Registrant.objects.get(id=self.request.session["registrant_id"])
         context["form_data"] = self.get_all_cleaned_data()
         context["registrant"] = registrant
+        if "category" in context["form_data"]:
+            context["form_data"]["category"] = dict(choices.location_choices)[
```
dsamojlenko, line 149: and then you'd just use `dict(location_choices)`

dsamojlenko (changes requested):
Looks great Nitya, a small suggestion for the placement of the location_choices dictionary

dsamojlenko (approved):

andrej-vasilj (approved):
Looks perfect to me

PR #411, opened 2021-03-04
Small change when running migrations

andrej-vasilj:
The --check flag was introduced in Django 3.1 so we can make use of it to help us detect situations where a migration was not successfully applied. Currently if a migration fails, the deployment will still succeed and bring up a new instance with bugs because the migrate command returns 0 even if it fails. With this flag, the migrate command will exit with status 1 so the deployment will fail.

No testing or action is needed for this.


dsamojlenko (approved):
Sweet

PR #413, opened 2021-03-08
Generate QR Code poster

dsamojlenko:
# Summary | Résumé

Generates the QR Code Poster. For now, the poster is not converted to PDF or sent by email. This PR uses an SVG template file to generate a visual representation of the poster, and renders it at a URL that is logged to the console. A future PR will convert this output to PDF and attach it to an email

When this is deployed, a new environment variable will need to be added with the Private Signing Key, and the Public Key will need to be shared with the Mobile App team for payload verification.

## Not included
- Doesn't convert to PDF
- Doesn't send the email
- Doesn't include terraform changes for environment variable (yet)
- This is not the final poster design, notably it does not render/print at the correct dimensions

## To test
- Be sure to follow the instructions in the next section to Create a signing key
- Go through the process of validating your email, then fill out the location details
- Once you hit "Make poster" on the Summary screen, you will see a URL output on the console. It will look like this: `http://localhost:8000/en/locations/7cb5422c-2c43-45f1-9882-1857cc283e3e/poster`
- Click the link to see the poster

## Create a signing key
You will need to set an environment variable `QRCODE_SIGNATURE_PRIVATE_KEY` which contains a private key for signing the QR Code payload. You can generate one for testing purposes.

From the command line enter a python shell. Import the utils file from the register app, and run generate_signature_key():

```
> python
Python 3.9.1 (default, Feb  3 2021, 07:38:02)
[Clang 12.0.0 (clang-1200.0.32.29)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from register import utils
>>> utils.generate_signature_key()
'g0xsyM4EJQqBD5zSPE6a4ORVq+wIpr3dWNx75q0otQQ='
```

Use the key output for the environment variable. Warning that this will not work with the Mobile App at this time as we will need to share the public key with the App team once we generate production keys.


register/utils.py
```diff
@@ -1,7 +1,76 @@
...
+    print(payload_bytes)
+
+    # Sign payload
+    signing_key = load_signature_key()
+    signed_b64 = signing_key.sign(payload_bytes, encoder=Base64Encoder)
+    print("Signed base64")
+    print(signed_b64)
+
+    # Build URL
+    url_prefix = "https://retrieval.wild-samphire.cdssandbox.xyz/exposure-configuration/download.html#"
```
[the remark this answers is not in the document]
dsamojlenko: nice!

register/tests.py
```diff
@@ -218,15 +247,75 @@ def test_location_model_generates_short_code_on_save(self):
...
+        code = utils.generate_random_key(5)
         self.assertEqual(len(code), 5)
 
     def test_generate_short_code_alphanumeric(self):
-        code = generate_random_key()
+        code = utils.generate_random_key()
         self.assertTrue(code.isalnum())
+
+    def test_generate_payload(self):
+        location = Location.objects.create(
```
andrej-vasilj, line 269: This is a bit noisy, I would recommend setting the location in the test setup function, that way it'll be available for all of the tests

register/tests.py
```diff
@@ -218,15 +247,75 @@ def test_location_model_generates_short_code_on_save(self):
...
+            address="Address line 1",
+            city="Ottawa",
+            province="ON",
+            postal_code="K1K 1K1",
+            contact_email="test@test.com",
+            contact_phone="613-555-5555",
+        )
+
+        qrcode = utils.get_signed_qrcode(location)
+        self.assertTrue(is_svg(qrcode))
```
andrej-vasilj, line 321: Is there anyway to test that it's actually signed not just b64 byte object? Not sure if that's important though so don't worry about it too much...

register/views.py
```diff
@@ -171,9 +172,33 @@ def done(self, form_list, form_dict, **kwargs):
...
 
+        poster_url = reverse("register:poster_view", kwargs={"pk": location.pk})
+        print("http://localhost:8000{url}".format(url=poster_url))
+        # Generate poster and send
+
         return HttpResponseRedirect(reverse("register:confirmation"))
 
 
+class PosterView(TemplateView):
+    template_name = "register/poster.svg"
```
andrej-vasilj, line 183: This line blew my mind!!!!! I never would have thought of this or that it's even possible to directly render to an svg as if it was an html! This made my day :)

Where did you get this idea???!

andrej-vasilj (approved):
Great PR Dave! This came together much simpler than I had envisioned actually! Looking really slick so far!

register/tests.py
dsamojlenko: good point - I can use the verify key to check it. will do.

register/views.py
dsamojlenko: Pretty happy with how this turned out ;)

sakura-sketch (commented):
Looks good! Simple and efficient

PR #417, opened 2021-03-09
Provincial detour page

sakura-sketch:
An unavailable detour page for provinces that aren't participated in the QR code registration


register/views.py
```diff
@@ -115,16 +115,29 @@ def get_success_url(self):
...
     "summary": "register/summary.html",
     "success": "register/success.html",
 }
 
+included_provinces = {"ON", "SK"}
+
+
+def check_for_province(wizard):
+    data = wizard.get_cleaned_data_for_step("address") or {}
+    print(included_provinces.__contains__(data.get("province")))
```
dsamojlenko, line 131: can you remove these print() debugging calls

dsamojlenko (changes requested):

register/forms.py
```diff
@@ -81,6 +81,14 @@ class RegisterSummaryForm(HealthcareBaseForm, forms.Form):
     pass
 
 
+class LocationUnavailableForm(HealthcareBaseForm, forms.Form):
```
andrej-vasilj, line 84: I don't think we need to explicitly declare empty forms here.

register/urls.py
```diff
@@ -8,12 +8,18 @@
...
 
+from register.views import check_for_province
+
 app_name = "register"
 
+
+# order matters here
 named_location_forms = (
     ("address", LocationAddressForm),
+    ("unavailable", LocationUnavailableForm),
```
andrej-vasilj, line 22: You could probably replace this with ("unavailable", forms.Form) and achieve the same effect...

register/views.py
```diff
@@ -123,8 +125,20 @@ def get_success_url(self):
 }
 
 
+def check_for_province(wizard):
+    data = wizard.get_cleaned_data_for_step("address") or {}
+    provinces = HealthcareProvince.objects.filter(qr_code_enabled=True)
+    provinces_values = provinces.values_list("abbr", flat=True)
+    provinces_list = list(provinces_values)
+
+    if data.get("province") is not None:
```
andrej-vasilj, line 134: I think it would suffice to use a simple Pythonic construct like:

return data.get('province') not in provinces_list

As a general piece of advice I would suggest to avoid using internal functions like '__contains__' if at all possible. Python doesn't strictly enforce class encapsulation, but the notation of using a single or double underscore indicates the intention to keep something private so it should generally be respected unless there's no other way around.

register/views.py
```diff
@@ -123,8 +125,20 @@ def get_success_url(self):
...
+        return not provinces_list.__contains__(data.get("province"))
+    return False
+
+
 class LocationWizard(NamedUrlSessionWizardView):
     def get_template_names(self):
+        print(self.steps.current)
```
andrej-vasilj, line 141: Probably don't need this print statement anymore?

andrej-vasilj (changes requested):
Hey, nice work! Looks like it should achieve the desired effect! Just have a few small comments..

register/views.py
sakura-sketch: Sure! Thank you so much!

register/views.py
sakura-sketch: Yeah, I missed that one

andrej-vasilj (changes requested):
Hey Nitya, thanks for the changes. I looked at this PR again and I still have some concerns. I see that your approach will work, but only superficially. I can reach the 'unavailable' page when testing this PR, but I can also just by-pass it by manually entering the next step in the browser url. Like I can just change 'unavailable' for 'category' and there's nothing blocking me from continuing. You'll still have to ensure that the next steps of the process are blocked if the province is not correct. I suggest that you take a look at how I did this for the OTK sms sending and apply the same approach.

andrej-vasilj (approved):
Sorry I didn't know that the intention was to still allow people to continue with the registration even after seeing the page. It looks like it serves the desired purpose given the new insights!

dsamojlenko (approved):

PR #435, opened 2021-03-11
Added more logic to adding/removing datetimes

andrej-vasilj:
Added some more logic to the back-end for adding/removing datetimes. There are some more conditions in place now to help with the edge cases:

- All datetimes need to be valid before a new one can be added
- Datetimes can be removed even if there are invalid ones
- All valid datetimes are cached when a new one is added/removed to help avoid continuity gaps
- Invalid datetimes will not be cached thus they will revert to the previously cached date or to empty

Feel free to test out those scenarios


smcmurtry (approved):
Tested and works great! Looks like you just need to `pipenv run format`

PR #439, opened 2021-03-12
Use the user's local timezone in alerting flow

### LINKED ISSUE #438: Use user's timezone for date times entered in alerting flow
Currently we are assuming the datetimes entered in the alerting flow (outbreaks/templates/datetime.html) are in UTC:
https://github.com/cds-snc/covid-alert-portal/blob/500e80e102569ba1bf422c91e588d6e7b214f90b/outbreaks/forms.py#L69
https://github.com/cds-snc/covid-alert-portal/blob/500e80e102569ba1bf422c91e588d6e7b214f90b/outbreaks/views.py#L299
https://github.com/cds-snc/covid-alert-portal/blob/97ba710968dea068368b48f301e7d4c600bd2a54/portal/settings.py#L222

They should be in the timezone the user's browser is set to. 

Here is some documentation from Django that could be helpful: https://docs.djangoproject.com/en/3.1/topics/i18n/timezones/

It is easy to get this info with client side js:
```
var now = new Date();
var timezoneOffset = now.getTimezoneOffset();
```

Although I'm not sure the best way to pass that back to the server - maybe a hidden form field on the datetime.html page? Not sure.

smcmurtry:
# Summary | Résumé

Closes #438. What I did:
- added some js to populate a hidden `<input>` with the user's browser's timezone offset
- added the hidden `<input>` to the datetime.html template
- verified that this input element is not visible to screenreaders (works the same way as the csrf input)
- sending this tz offset to the server so the correct utc timestamp is generated
- added this offset to the user's session so it can be used to display the saved utc time as a local time on the "confirm" and "confirmed" pages

### DIFF
--- a/outbreaks/forms.py
+++ b/outbreaks/forms.py
@@ -67,6 +67,8 @@ class DateForm(HealthcareBaseForm):
 
     def get_valid_date(self, data, i):
         tz = pytz.timezone(settings.TIME_ZONE or "UTC")
+        print("settings.TIME_ZONE", settings.TIME_ZONE)
+        print("tz", tz)
         return datetime(
             year=int(data.get(f"year_{i}", -1)),
             month=int(data.get(f"month_{i}", -1)),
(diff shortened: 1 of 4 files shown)


dsamojlenko:
Closing as this work is being incorporated into the redesigned section

PR #443, opened 2021-03-17
Remove line numbers from translation file

dsamojlenko:
# Summary | Résumé

This update removes line numbers from our translations .po file, and updates the command reference in the readme with two new arguments `--add-location=file` and `--no-wrap`.

## TL/DR

From now on, when running the `makemessages` command to capture content changes, use the new command arguments:

```
python manage.py makemessages -l fr --add-location=file --no-wrap
```

## Background

When there are a lot of inflight PRs, managing merge conflicts in the translations file can get a bit hairy. 

One reason for is that the `django.po` file can end up with a lot of changes even if no text has changed- if template code has been moved around then you can end up with a cascade of line number updates throughout the .po file.

Adding the `--add-location=file` argument means that we will continue to write out the file location of the string (helpful) but not the line number.

The second new argument `--no-wrap` will prevent long lines from being split into multiple lines. This is intended to make the files more readable, but in practice makes them more prone to conflict/editing errors.


omartehsin1 (approved):
Perfect. Will help reduce merge conflicts!

sakura-sketch (approved):

PR #448, opened 2021-03-18
added created_at field for location and registrant tables

sakura-sketch:
an extra column of created_at timestamp has been added to location and registrant tables. If you have existing data then the current date and time will be added on migration


dsamojlenko (approved):
Look great!

PR #458, opened 2021-03-26
Send Email Again button

omartehsin1:
# Summary | Résumé

This branch takes care of the send email button. Changed the button to a hyperlink that takes the user back to the email page with the field populated with the entered email address.

Changes made:
- Content / translation changes
- Updated button design.


register/views.py
```diff
@@ -253,3 +253,7 @@ def get_success_url(self):
 
 class QRSupportPageView(TemplateView):
     template_name = "register/qr_support_page.html"
+
+
+class ResendEmailView(TemplateView):
+    template_name = "register/resend_email.html"
```
dsamojlenko, line 259: Since we don't need to manipulate data or anything before rendering the view, we don't need this Class and can just return the view from the URL definition (see second comment on the urls file)

register/urls.py
```diff
@@ -67,4 +67,5 @@
         TemplateView.as_view(template_name="register/success.html"),
         name="success",
     ),
+    path("resendemail", views.ResendEmailView.as_view(), name="resendemail"),
```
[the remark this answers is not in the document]
dsamojlenko: Also, we can just return the view from here instead of using a Class-based view. ie:

`path("resendemail", TemplateView.as_view(template_name="register/resend_email.html"), name="resendemail"),`

register/templates/register/registrant_email_submitted.html
```diff
@@ -19,10 +19,7 @@ <h1>{% trans "Confirm your email address" %}</h1>
     <p>We sent a link to {{ email }}.</p>
   {% endblocktrans %}
 
-  <p>{% trans "If you did not get the email, check your junk or spam folder." %}
-
-  <div class="content-button">
-    <a href="{% url 'register:registrant_email' %}" role="button" draggable="false">{% trans "Send email again" %}</a>
-  </div>
+  <p>{% trans "Check your inbox, junk or spam folder for an email with the subject line <q>Confirm Email for your poster</q>" %}</p>
+  <p>{% trans "If you did not get the email, we can " %} <a href="{% url 'register:registrant_email' %}">{% trans "send it again" %}</a></p>
```
dsamojlenko, line 23: Probably consider using blocktranslate with a placeholder here instead of partial strings: https://docs.djangoproject.com/en/3.1/topics/i18n/translation/#blocktranslate-template-tag

register/views.py
```diff
@@ -66,7 +77,6 @@ def dispatch(self, request, *args, **kwargs):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context["submitted_email"] = self.request.session.get("submitted_email")
-        del self.request.session["submitted_email"]
```
dsamojlenko, line 69: oh yeah

dsamojlenko (approved):

PR #459, opened 2021-03-29
feat: adds a11y tracking to CI pipeline

maxneuvians:
This PR adds a call to the a11y tracking tools into the CI pipeline. After both the portal and the qrcode components have been deployed it will make an API call to the a11y tracking tools API that request a spidering of the public facing pages of each component.


dsamojlenko (approved):
This is awesome!

PR #460, opened 2021-03-31
added noscript tag when javascript is disabled

sakura-sketch:
Address line 1 in the registration site is now visible when javascript is disabled


dsamojlenko (approved):
Looks good, pushed a small related bugfix

PR #464, opened 2021-04-01
Update the canada post trial api key

dsamojlenko:
# Summary | Résumé

We are currently using a Trial account of the Canada Post API. The Trial account is only good for 14 days. So every 14 days, we need to create a new Trial account and create a new API key. Whenever procurement gets sorted out, this problem will go away.

Previous related: #447 #386


register/static/register/js/address-autocomplete.js
```diff
@@ -3,7 +3,7 @@
  * but it'll be locked to our domain and is not sensitive
  * https://www.canadapost.ca/pca/support/kb/address-complete-basics/#safe
  */
-const key = 'XF51-CY97-JE96-HK81'
+const key = 'RY99-ZY65-YP98-AT12'
```
dorianjp, line 6: this is an interesting situation w procurement

What do you think about putting this into the TF env? we'll need to w the production key regardless, and for the time being we could flip the variable via kubectl in the cli

dorianjp (approved):
just a comment for next time

register/static/register/js/address-autocomplete.js
dsamojlenko: It's not a sensitive key, so no need to store in secrets ... Canada Post recommends just locking the key to your domain(s) and not worrying about leakage.

PR #478, opened 2021-04-12
Bugfix/confirmation page redirect

dsamojlenko:
# Summary | Résumé

Fixes a bug where a user that has a verified email, but hasn't completed the Location wizard could still access the Confirmation screen which would cause an error.

Also added/updated some tests around this section, and improved the Location Detour test which was calling another test to setup its required state.

## To test
- Verify your email
- Instead of filling out the form, go directly to `/en/confirmation`
- You should be redirected back to the start of the form
- Also, have a look at the tests and run them


omartehsin1 (approved):
Looks good!

PR #482, opened 2021-04-15
Updating API key to new trial account

dsamojlenko:
# Summary | Résumé

Just updating the temp api key again


dorianjp (approved):
this is fine

PR #483, opened 2021-04-15
Refactor - Create base form template for more re-usability

dsamojlenko:
# Summary | Résumé

This small refactor creates a cds_form_base.html which contains the basics of a form template which can be extended to customize how fields are rendered for particular forms.

So, for example currently all of our forms include the cds_form.html form template. This works for 90% of our forms, but recently our Location Contact form needed some custom rendering of the phone number/extension fields. 

In this case, we created a new form template, basically repeating everything in the cds_form.html template, and customizing the two fields we wanted to customize.

In another upcoming change to the location_category form, we'll need some more control over the rendering again. So rather than continually copying the cds_form.html and having all that duplicate code, this breaks that form template up so that our custom templates can extend the basic form stuff, and then just customize the field rendering.


dorianjp (commented):
It's good to have a way to escape the default structure when needed.

One other option which I believe follows your intent would be to wrap each block in cds_form.html with various replaceable blocks (content, submit, javascript, etc), but remaining the content within the one template.

Any other form could still extend the cds_form.html, but overwrite individual blocks as needed, and existing forms wouldn't be impacted in much the same you have here.

dorianjp (approved):
nice work

PR #484, opened 2021-04-15
Fix/adjust input field widths and display

dsamojlenko:
# Summary | Résumé

Fixes a few input styling/sizing issues, see screenshots below.

Also includes a minor change to how the conditional validation is handled on the Category step.

## Screenshots

| Something else | Postal code |
| ----- | ----- |
| image | image

Also improves the mobile/desktop styling on the Extension field on the Contact form slightly

| Desktop | Mobile |
| ----- | ----- |
| image | image |


register/templates/register/includes/category_form.html
```diff
@@ -0,0 +1,42 @@
...
+            {% endif %}
+            {% with "aria-describedby:error--"|add:field.html_name as describedBy %}
+                {{ field|add_error_attr:describedBy  }}
+            {% endwith %}
+        </div>        
+      {% endif %}
+    {% endfor %}
+  </div>
+
+  <div class="fieldWrapper--container{% if form.non_field_errors %} fieldWrapper--container--has-error{% endif %} category-other">
```
dorianjp, line 27: If the only reason for the custom rendering is applying a class to a field, we have options to add additional class(es) to fields in the forms.py which will be applied to the field input.

`category_description = forms.CharField(
        label=_("Tell us the type of establishment."), required=False,
        widget=forms.TextInput(attrs={'class':'category-other'})
    )`
    
    https://docs.djangoproject.com/en/3.2/ref/forms/widgets/#django.forms.Widget.attrs

But this works too and gives you more flexibility beyond applying one class to the input element itself

register/static/register/scss/_custom.scss
```diff
@@ -134,4 +139,15 @@ body.qrcode {
             }
         }
     }
+
+    .category-other {
+        border-left: 4px solid #C4C4C4;
+        margin-left: 18px;
+        padding-left: 20px;
+        max-width: 450px !important;
```
dorianjp, line 147: are there cases the max-width is overridden? I tried a test without !important and it seemed ok.

register/templates/register/includes/category_form.html
dsamojlenko: Yeah, I was trying to solve it like that with field attrs, but that only makes the class available on the input, and I needed to apply it to an outer element, couldn't figure out how to get a class onto that outer fieldWrapper--container.

register/static/register/scss/_custom.scss
dsamojlenko: may have been one of those "something's not working, throw-spaghetti at the wall to see what sticks" moments... you're right, not necessary.

dorianjp (approved):
lgtm

PR #487, opened 2021-04-16
Address Autocomplete support for Internet Explorer 11 

dsamojlenko:
# Summary | Résumé

The address autocomplete was breaking on IE11. The underlying accessible-autocomplete component from GDS supports IE11, so it turned out the problem was our implementation. We used a few arrow functions and Array.find() method, neither of which are supported in IE11. 

This PR refactors the code to use regular functions and Array.filter().


dorianjp (approved):
nice that you could isolate the change without using a polyfill / shim, though those are handy when the alternative is a big refactor

PR #488, opened 2021-04-19
Add location address to poster email subject

dsamojlenko:
# Summary | Résumé

Adds location_address to the notify payload for the poster email. Notify template already updated.


dorianjp (approved):
lgtm

PR #490, opened 2021-04-20
Set default auto field to stop Django 3.2 complaining

dsamojlenko:
# Summary | Résumé

Since upgrading to Django 3.2, the console has been barfing a bunch of warnings related to setting a default for Automatically created Primary Keys. This sets a global default to the current behaviour (AutoField) to prevent future unwanted migrations when a future version of Django changes the default to BigAutoField.

[See documentation](https://docs.djangoproject.com/en/3.2/releases/3.2/#customizing-type-of-auto-created-primary-keys)


dorianjp (approved):
lgtm

PR #497, opened 2021-04-26
Add data length validation to text inputs

dsamojlenko:
# Summary | Résumé

Adds some max_length properties to the text inputs, previously they would accept any length and presumably would truncate or error when attempting to insert strings longer than the model spec.

Also, there was a "OtherFieldInput" form class that wasn't being used, so cleaned it up.


dorianjp (approved):
lgtm

PR #498, opened 2021-04-26
Set custom validation message for invalid phone number

dsamojlenko:
# Summary | Résumé

Changes the validation message for invalid phone numbers from `+12125552368` to something that looks more like a phone number `613-555-5555`

Will also now translate that message as it was not translated before.


dorianjp (approved):
lgtm

PR #500, opened 2021-04-26
Expand searchable params on Location search

dsamojlenko:
# Summary | Résumé

Make locations more findable by using PostgreSQL full-text search features. Since the database features are Postgres-only, also updated the Github Actions to use a PostgreSQL service.

Note that in local development, search will not work if using the default sqlite database. You should start up a postgres service and set the DATABASE_URL environment variable.


outbreaks/views.py
```diff
@@ -41,25 +42,28 @@ class SearchView(PermissionRequiredMixin, Is2FAMixin, ListView):
     paginate_by = 5
     model = Location
     template_name = "search.html"
+    search_fields = ["name", "address", "city", "postal_code"]
```
dorianjp, line 45: was this left from the previous search?

dsamojlenko: Yup, should have gotten an unused variable error... weird. cleaned it up!

dorianjp (approved):
nice on bringing up the changes to the CI as well

PR #503, opened 2021-04-28
Fix localization on profile page plus add browser tz support to request

dorianjp:
Trello:
https://trello.com/c/gG8YZyrV/768-bugfix-date-localization-for-password-last-updated-on-profile-view-does-not-work-on-the-production-server

Steps to reproduce error:
* Login to portal
* Click manage account
* Error should be evident if password has been changed > 0 times

Fix:
* Refactor to more appropriate tz localization
* Add support for local browser tz as a cookie
* Maintain ability to get local time in middleware

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - QR Code Registration form with Email verification
 - Location search with Canada Post Address Complete API
 
+## [1.19.4] - 2021-04-28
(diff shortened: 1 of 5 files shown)


dorianjp:
[Bugfix: Date Localization for Password Last Updated on Profile View does not work on the production server](https://trello.com/c/gG8YZyrV/768-bugfix-date-localization-for-password-last-updated-on-profile-view-does-not-work-on-the-production-server)

dsamojlenko (approved):

PR #505, opened 2021-04-28
Release 1.19.4

dsamojlenko:
# Summary | Résumé

This minor version includes:
- Bugfix for localized Password Changed timestamp on Profile pages
- Render the Password Changed timestamp in the browser-reported local timezone


dorianjp (approved):
let it ride!

PR #508, opened 2021-04-30
Search improvements: Partial word search and ignore accents

dsamojlenko:
# Summary | Résumé

Improvements to the Location search rolled out in #500 

The Django full-text search function wrappers around the Postgres full-text search features are limited in that they do not support partial word search syntax. We also had an issue with french accents not being ignored for search purposes.

This PR refactors the search code to:
- Dynamically build up the search query with appropriate wildcard markup
- Use the underlying Postgres functions directly
- Ignore french accents for the purpose of search, so "Montreal" will return results including "Montréal"

A search query containing three words: 
- `first second third` 

would result in the following search query: 
- `first:* & second:* & third:*`

So, each string part is treated as a starts_with wildcard search, and are combined with AND operators.

[Based on idea described here](https://www.fusionbox.com/blog/detail/partial-word-search-with-postgres-full-text-search-in-django/632/)

## To test
Before checking out this branch, try some partial word searches and note that there are no results. Also if you have a location with the city "Montréal", try searching for "Montreal" (no accent) and note there are no results.

Then checkout this branch and try both of those searches again.

## Important note
To ignore accents, had to enable an Unaccent Postgres extension through a migration. I am uncertain if this migration will actually run on Staging/Production. Something to watch for when merging to Staging.


outbreaks/views.py
```diff
@@ -46,14 +64,29 @@ class SearchView(PermissionRequiredMixin, Is2FAMixin, ListView):
...
-                Location.objects.annotate(
-                    search=SearchVector("name", "address", "city", "postal_code")
-                )
-                .filter(search=SearchQuery(searchStr))
-                .order_by("name")
+            #
+            # Also note: this was originally based on Django built-in functions,
+            # but they don't support partial string search so refactored
+            # to use postgres directly, see:
+            # https://www.fusionbox.com/blog/detail/partial-word-search-with-postgres-full-text-search-in-django/632/
```
dorianjp, line 73: Nice approach, but would adding the param search_type to the [SearchQuery be similar to using the raw tsvector ](https://docs.djangoproject.com/en/3.2/ref/contrib/postgres/search/#searchquery)approach here after you added the operators?


Also did you test for what happens if the user inserts their own operators?

dsamojlenko: I wasn't able to get things working using search_type and standard syntax with the processed query params. I can get it sorta working (inconsistently), but then I can't use the postgres unaccent function which helps with french-accented characters, and I'm not sure there's another solution for that.

The process_query function strips any existing operators or funny characters, so they don't affect the results.

dorianjp (approved):
nice work!

PR #510, opened 2021-05-04
Update canada post trial api key

dsamojlenko:
# Summary | Résumé

Just updating to another trial account for the last time


dorianjp (approved):
lgtm

PR #514, opened 2021-05-04
Gracefully fallback if an error comes back from the Canada Post api

dsamojlenko:
# Summary | Résumé

If an error comes back from the Canada Post API (for example, when credits run out), just display "No results" and allow the user to continue to fill in the form manually.


dorianjp (approved):
Do we have any way to be alerted if this fails / lapses billing etc?

dsamojlenko:
> Do we have any way to be alerted if this fails / lapses billing etc?

Yeah, there's a dashboard for the Production account, I'll get some accounts setup for team members who need to keep an eye on it.

PR #519, opened 2021-05-07
Use list_display and model properties for csv output

dsamojlenko:
# Summary | Résumé

The previous version of CSV output would only render fields listed in the originating model Meta, and so wouldn't include any related properties that were added into the Admin list view.

This changes the mixin to use the list of properties from the Admin view `self.list_display` so that we are dumping out what the user sees onscreen.

This requires that everything in the list_display is a Model property, so had to convert one admin view property to model propoerty.

## To test
Visit the Admin section Locations, Registrants, or Notifications, and try exporting data. The CSV will include the same data you see on screen, including model properties (note specifically Locations->registrant_email and registrant_id)

### DIFF
--- a/outbreaks/admin.py
+++ b/outbreaks/admin.py
@@ -14,14 +14,11 @@ class NotificationAdmin(admin.ModelAdmin, ExportCsvMixin):
         "created_by",
         "severity",
         "location",
-        "short_code",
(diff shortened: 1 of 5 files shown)


dorianjp (approved):
lgtm

PR #504, opened 2021-04-28
Feature/admin panel improvements and Location/Outbreak Summaries

dsamojlenko:
# Summary | Résumé

Improves some of the registration/outbreaks related Admin panels, and adds a Location Summary panel and a Notifications Summary panel. This is the start of our metrics dashboard work - surfacing service-level metrics that can be extracted from the database. This first iteration is just to get a sample of what's possible, can do more refining with feedback.

## Location Summary
- No. of posters printed per Province
- No. of posters printed per City

## Notifications Summary
- No. of Outbreaks sent per Province
- No. of Outbreaks sent per City
- No. of Outbreaks sent per Location

## Registrants
- Added inlined Locations list for each Registrant

## Locations
- Added a bunch of filtering options to the main listing page (province/city/created date)

PR #520, opened 2021-05-07
Auto-generated migrations for summary models

dsamojlenko:
# Summary | Résumé

These migrations were auto-generated as a result of the Summary models that were added in #504


dorianjp (approved):
sure

PR #528, opened 2021-05-13
Add missing translation

dsamojlenko:
# Summary | Résumé

Adds a missing translation for the BETA banner


dorianjp (approved):
looks like French to me

PR #530, opened 2021-05-14
Exclude status from basicauth middleware

dsamojlenko:
# Summary | Résumé

Skip basicauth on /status for DNS health checks


dorianjp (approved):
a simple fix, nice

PR #536, opened 2021-05-18
Enforce a character limit that doesn't run off the poster page

dsamojlenko:
# Summary | Résumé

In QA testing, an issue was identified that long business/location names would run off the poster as there is a hard limit to the number of characters that can fit, and wrapping is not an option.

Testing with the widest character on the keyboard, the M or W, the maximum number of characters that can comfortably appear is approximately 36.

36 is a bit tight, and no business/location has only Ws in their name, so compromising at 65.


dorianjp (commented):
easy,
do we want to also include messaging advising why we chose this limit as opposed to the standard ## max message? Yedida may want to weigh in.

dsamojlenko:
> easy,
> do we want to also include messaging advising why we chose this limit as opposed to the standard ## max message? Yedida may want to weigh in.

Good point

locale/fr/LC_MESSAGES/django.po
```diff
@@ -2145,6 +2145,11 @@ msgstr "Autre"
 msgid "Tell us the type of establishment."
 msgstr "Quel est le type de votre établissement?"
 
+#: register/forms.py
+#, python-format
+msgid "Your name is longer than the %(limit_value)d character limit"
+msgstr ""
```
dsamojlenko, line 2151: @EliseKa svp

EliseKa:
```suggestion
msgid "Your name is longer than the %(limit_value)d character limit"
msgstr "Le nom dépasse la limite de %(limit_value)d caractères"
```

dsamojlenko:
Updated error messaging, ready for re-review

register/forms.py
```diff
@@ -77,7 +77,15 @@ def clean(self):
 
 
 class LocationNameForm(HealthcareBaseForm, forms.Form):
-    name = forms.CharField(label="", max_length=200)
+    name = forms.CharField(
+        label="",
+        max_length=65,
+        error_messages={
+            "max_length": _(
+                "Your name is longer than the %(limit_value)d character limit."
```
dorianjp, line 85: There's a period here but not in the PO file.

dsamojlenko: fixed

dorianjp (approved):
lgtm

PR #537, opened 2021-05-18
Set the default format for Phone numbers when displaying from db

dsamojlenko:
# Summary | Résumé

This was identified during QA testing of the Registration site. We use a PhoneNumber django extension for validating/storing phone numbers. Currently when displaying numbers, we just pull them from the database and display them, so they look like: `+16135555555`. 

Note: The "global" setting `PHONENUMBER_DEFAULT_FORMAT` is currently only being applied to the Register site. This is due to the fact that there are a few places on the Portal side where the new format breaks things (tests/urls), so until we can do further testing and fixing on that side, going to just restrict this to the registration site.


dorianjp (approved):
handy library

PR #538, opened 2021-05-18
Validate Canadian Postal code

dsamojlenko:
# Summary | Résumé

Currently the Address form will accept anything up to 10 characters for a Postal Code. This just enforces proper postal code format.

To do this, the [Django localflavor package for Canada](https://django-localflavor.readthedocs.io/en/latest/localflavor/ca/) was added.

Note: we are not able to validate that the Postal code entered is the correct one for the Address, but we can at least ensure that the Postal code is correctly formatted and valid on its own.


register/forms.py
```diff
@@ -107,7 +108,7 @@ class LocationAddressForm(HealthcareBaseForm, forms.Form):
     )
     city = forms.CharField(label=_("City"), max_length=100)
     province = forms.ChoiceField(label=_("Province or territory"), choices=provinces)
-    postal_code = forms.CharField(label=_("Postal code"), max_length=10)
+    postal_code = CAPostalCodeField(label=_("Postal code"), max_length=10)
```
dorianjp, line 111: Do we want to add any hinting ahead, it looks like it's a hard regex on:
        'invalid': _('Enter a postal code in the format XXX XXX.'),

locale/fr/LC_MESSAGES/django.po
```diff
@@ -2221,6 +2221,10 @@ msgstr "Province ou territoire"
 msgid "Postal code"
 msgstr "Code postal"
 
+#: register/forms.py
+msgid "Enter a valid Canadian postal code."
+msgstr ""
```
dsamojlenko, line 2226: @EliseKa

EliseKa:
```suggestion
msgid "Enter a valid Canadian postal code."
msgstr "Entrez un code postal canadien valide."
```

register/forms.py
dsamojlenko: Got an updated validation error from Yedida

dsamojlenko:
Updated the error message, ready for re-review

dorianjp (approved):
lgtm

PR #543, opened 2021-05-20
Add full-text search to Alert history

dsamojlenko:
# Summary | Résumé

Switches out the query syntax for full-text search so the results match the main Location search.


outbreaks/views.py
```diff
@@ -536,30 +536,30 @@ def get_context_data(self, *args, **kwargs):
...
-                qs = Notification.objects.all()
-            else:
-                qs = Notification.objects.filter(location__province=province)
+            qs = Notification.objects.filter(location__province=province)
+
+        if search:
+            query = process_query(search)
+
+            qs = qs.extra(
+                tables=["register_location", "outbreaks_notification"],
```
dsamojlenko, line 550: this was the bit that took some figuring out ... I tried setting up the join on the default queryset with `select_related('register_location')` but kept getting an error about a missing table in the WHERE clause, even though `print(qs.query)` would output a perfectly valid SQL query that I could run in TablePlus. 

Specifying the tables here fixed it 🤷

dorianjp (approved):
lgtm

PR #544, opened 2021-05-25
Update the "shows on poster" label with the correct highlight colour

dsamojlenko:
# Summary | Résumé

| Before | After |
| ----- | ----- |
image | image


dorianjp (approved):
magnificent

PR #545, opened 2021-05-25
Add regex validation to prevent alpha chars in phone number field

dsamojlenko:
# Summary | Résumé

The PhoneNumberField widget we're using does a great job of validating/formatting phone numbers, but had an odd "feature": You could enter alpha characters in the phone number field, and it would convert them to corresponding numbers on a phone keypad. Interesting feature, but possible not expected behaviour. 

This PR adds a little regex to check for alpha characters in the Phone number field and throws a validation error.

PR #550, opened 2021-05-26
Add regex validator to disallow funny characters 🥸

dsamojlenko:
# Summary | Résumé

🥸 keep this guy out of our CharFields


locale/fr/LC_MESSAGES/django.po
```diff
@@ -2262,6 +2261,10 @@ msgstr "Autre."
 msgid "Tell us the type of place."
 msgstr "De quel type de lieu s’agit-il?"
 
+#: register/forms.py
+msgid "Only enter letters or numbers."
+msgstr ""
+
```
dsamojlenko, line 2267: @EliseKa svp

EliseKa:
```suggestion
#: register/forms.py
msgid "Only enter letters or numbers."
msgstr "Vous ne pouvez saisir que des lettres et des chiffres."

```

locale/fr/LC_MESSAGES/django.po
```diff
@@ -2218,6 +2211,12 @@ msgstr "Votre session a expiré. Reconnectez-vous pour continuer à utiliser le
 msgid "— Scan a place registration"
 msgstr ""
 
+#: register/forms.py
+#, fuzzy
```
dorianjp, line 2215: going to need to watch this vs the fix in the other pr #545

register/forms.py
```diff
@@ -86,10 +87,16 @@ def clean(self):
             raise forms.ValidationError(_("Tell us the type of place."))
 
 
+alphanum_validator = RegexValidator(
+    r'^[0-9A-zÀ-ÿ-_\s,.!?"\'\(\)]*$', _("Only enter letters or numbers.")
```
dorianjp, line 91: I checked a few strings and noticed we block a fair bit of other punctuation / symbols like # or :, how open or closed do we want to be with this?

dsamojlenko: Yeah, let's open it up to discussion. I started with alphanumeric, but added accented characters and a few random "special" chars, but not sure if we want to be more/less permissive

dorianjp (approved):
thanks for the update

PR #557, opened 2021-05-27
Add Canada geo-restriction

dsamojlenko:
# Summary | Résumé

Restrict access to Canada IPs only

## Also included
- Identified a few Cloudwatch alarms that were named the same as the Portal alarms, which was causing some problems. Renamed the affected alarms.
- ~~Added some tfsec ignores and fixes for sec scan issues.~~ These ended up in another PR


dorianjp (approved):
worth a try!

PR #560, opened 2021-05-31
Remove basicauth from Production

dsamojlenko:
# Summary | Résumé

Basicauth was meant as a temporary measure on the Production registration site. This removes it in preparation for launch.


dorianjp (commented):
@dsamojlenko 
Do we also want to get rid of the override middleware class as well?

dsamojlenko:
> @dsamojlenko
> Do we also want to get rid of the override middleware class as well?

Good catch, removed it.

dorianjp (approved):
lgtm

dsamojlenko:
Closing this as we will be decommissioning the Registration site in Production.

PR #561, opened 2021-05-31
Fix datepicker on IE

dsamojlenko:
# Summary | Résumé

Internet Explorer is still a thing sometimes, and it can't iterate over a nodelist, which is what is returned by querySelectorAll. This changes the nodelist to an array so the forEach iterator works on IE.


dorianjp (approved):
let IE rest in pieces

PR #576, opened 2021-06-02
Update dependencies

dsamojlenko:
# Summary | Résumé

In light of the recent Dependabot PR #575 figured we could just do a package bump for everything.


dorianjp (approved):
let it ride!

PR #579, opened 2021-06-03
Create JMeter Load Testing Test Plan

dsamojlenko:
# Summary | Résumé

JMeter test plan for Registration Site load test.

There are two plans: 

## Registrant email confirmation
Renders the email page, submits an email.

## Location registration
Steps through the Registration steps, beginning after email confirmation and ending with generating and downloading the posters. 

To run this plan, you will need to manually confirm an email and grab the Django session id from the browser cookies, and add it to User defined variables (see below)

## User defined variables
To use these scripts, you will need to configure some variables in the User defined variables section:
- host: the base url (defaults to localhost)
- scheme: http/https
- testEmail: the email you want to submit in the confirmation step
- port: if https, should be 443. if local, should be 8000
- sessionCookie: for the Location registration plan, 

## Use caution - you will send email!
When run, these scripts will end up filling out forms and submitting them in whatever environment you choose. If running against staging/production, that means you will definitely be sending emails as part of the test. If running a load test, you could be sending thousands of emails. 

We need to discuss how to suspend the actual sending of email during a load test.

### DIFF
--- a/config/load_testing/Test Plan.jmx
+++ b/config/load_testing/Test Plan.jmx
@@ -1,0 +1,1401 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<jmeterTestPlan version="1.2" properties="5.0" jmeter="5.4.1">
+  <hashTree>
+    <TestPlan guiclass="TestPlanGui" testclass="TestPlan" testname="Test Plan" enabled="true">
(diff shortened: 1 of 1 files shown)


dsamojlenko:
closing as this is no longer needed

PR #580, opened 2021-06-03
Update the CP key again

dsamojlenko:
# Summary | Résumé

Switching back to a test key temporarily to sort out an issue with the prod key (see trello)


dorianjp (approved):
ok

PR #621, opened 2021-06-16
Make registrants editable

dsamojlenko:
# Summary | Résumé

In order to conduct load testing without sending a bunch of emails, make the Registrant table editable in the Admin panel so that we can change the registrant email to a test email from amazon that will return a success response without actually sending anything.


dorianjp (approved):
Free the table!

PR #623, opened 2021-06-17
enable autoscaling on staging for load testing

dsamojlenko:
# Summary | Résumé

This will enable autoscaling on Staging so that we can run our Load tests against it and hopefully see the infrastructure scale up.


dorianjp (approved):
lgtm

PR #635, opened 2021-06-22
Update the WAF config to match production

dsamojlenko:
# Summary | Résumé

Brings the WAF config inline with production:
- Create separate ACLs for Registration Site and Portal
- Separate WAF logging
- Separate WAF logs bucket


patheard (approved):
Looks good 👍🏻

PR #643, opened 2021-06-23
Add uniqueness constraint to Location model

dsamojlenko:
# Summary | Résumé

Saw a bunch of errors on Staging like this:

image

Figured out this is due to a race condition in the get_or_create used when creating a Location:

image

The intention is to retrieve an existing location if one matches the criteria and update, or create a new. But there is a race condition where two processes can get past the get() check and then both create the object.

The reason we're seeing this now is due to the load testing - one of the test runs obviously created duplicate entries under this condition.

This PR adds a uniqueness constraint to the model to prevent the duplicate from being created.


sylviamclaughlin (approved):
Looks good!

dsamojlenko:
Thanks for the review. Putting this in Draft for the moment, going to hold it back until I can fix the data on Staging otherwise this will fail spectacularly.

dorianjp:
We'd also probably want to do an iexact on the fields for case insensitivity if we go ahead with this. But it may not be necessary, as discussed.

dsamojlenko:
Going to close this for now. Don't think it's going to be necessary.

PR #645, opened 2021-06-23
Return an empty string on missing registrant model

dsamojlenko:
# Summary | Résumé

This shouldn't be possible, but somehow on Staging a Registrant has been deleted, orphaning its Locations. This is now causing an error on the Admin/Locations view because of the missing model - the Location model tries to populate registrant_email from the related model which no longer exists.

This just returns an empty string in that case.

### DIFF
--- a/register/models.py
+++ b/register/models.py
@@ -54,7 +54,7 @@ class Location(models.Model):
 
     @property
     def registrant_email(self):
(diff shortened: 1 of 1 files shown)


sylviamclaughlin (approved):
Looks good!

PR #646, opened 2021-06-23
Add cookie to disable sending poster email for testing

dsamojlenko:
# Summary | Résumé

To assist with load testing, we can set a cookie to prevent sending the poster email.

In our current testing scenario, we want to just test our application under load, not its external dependencies. This should be removed later.

## How to use it

If you don't want to send emails, set a cookie called `nonotify` with any text as contents. Ensure it's set on the domain you're testing against.


dorianjp (approved):
let it ride

PR #650, opened 2021-06-24
add location to ensure join on query when sorting by date

dorianjp:
# Summary | Résumé

Trello
https://trello.com/c/x9hELaXd/831-sort-search-results-on-the-alerts-history

Join was missing from generated SQL when sorting by date only, added location->name to force inner join when sorting by date.

Sort by date on alert history, search results should remain as expected with no additional entries.

### DIFF
--- a/outbreaks/views.py
+++ b/outbreaks/views.py
@@ -607,7 +607,9 @@ class HistoryView(SearchListBaseView):
             return qs.order_by(col if order == "asc" else col.desc())
         else:
             col = "start_date"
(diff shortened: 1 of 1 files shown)


dorianjp:
[Sort search results on the alerts history](https://trello.com/c/x9hELaXd/831-sort-search-results-on-the-alerts-history)

sylviamclaughlin (approved):
Tested it out and works as you would expect now.

dsamojlenko (approved):

PR #652, opened 2021-06-24
Add WhiteNoise for serving static assets + github actions refactor

dsamojlenko:
# Summary | Résumé

Adds [WhiteNoise](http://whitenoise.evans.io/en/stable/index.html) for serving static assets more efficiently. Hopefully this will help a bit with container performance.

Also added our registration site URLs to the CSP config, not sure if this had any effect.

Had to do a little work to get the tests to run, and in that process decided to refactor the Github Actions a bit:
- Installed python directly in the container rather than use the pipenv action we were using as this is recommended as a faster approach
- Combined the Code Coverage workflow into the code section of the main workflow
- Added dependency caching to speed things up a bit
- Run `collectstatic` to get tests passing

This has definitely sped up our CI. Not including LGTM which takes forever... the rest of our tasks went from roughly 11 minutes down to 5.


dorianjp (approved):
lgtm

PR #655, opened 2021-06-25
feat: split portal/qrcode scaling min/max capacity

patheard:
# Summary 
This allows the portal and QR code services to independently set
minimum and maximum scale capacities.

Also sets scale-out cool down back to 60 seconds as the CPU
metric is only captured every 60 seconds, so having a value
lower than this causes inconsistent scaling.

# Changes
* Update autoscaling targets with lowered CPU threshold.
* Update QR code autoscaling target with new min capacity of `4`.
* Recreate the QR code WAF POST limit rule (I'll take care of removing after the `tf apply`).


dsamojlenko (approved):
good idea

Issue #604, opened 2021-06-11
Please change English heading for first screen of OTK flow


YedidaZalik:
https://staging.covid-hcportal.cdssandbox.xyz/en/start/  
Please change English heading to

**Now you can give one-time keys**

sylviamclaughlin:
@YedidaZalik - I think there might be some discrepancies with another ticket. Should this be "Now you can give one-time keys" or "You are logged in to give one-time keys". Thank you so much for the clarification!

YedidaZalik:
@sylviamclaughlin my understanding is that there will be 2 different versions of this flow:

-  Users who are authorized to give one-time-keys **and** send QR alerts will log in to this screen: https://www.figma.com/file/znPqLTQQDcUkaNjdDcShiz/Exposure-Notification---Health-Portal?node-id=8334%3A62464
If on that screen they choose "Give one-time-keys", then the next screen should have the heading "Now you can give one-time keys"

But, if, they are only authorized to give one-time-keys, then they'll be taken directly to the one-time-key flow when they log in. In that situation, the heading should be "You're logged in to give one-time keys".

Thanks!

sylviamclaughlin:
@YedidaZalik - Thank you so much for the explanation! This makes it clearer.

sylviamclaughlin:
This has been approved and merged to stage.

PR #659, opened 2021-06-28
Modifying OTK-flow as per git issue #604

sylviamclaughlin:
# Summary | Résumé

Modifying otk-flow as specified in github issue #604 

## Test instructions | Instructions pour tester la modification
Testing this feature can be a little bit tricky. There could have been a better way to do it (ie modifying the request object using the developer tools), but I wanted to replicate exactly how the flow will be executed.

1. To test the "Now you can you give one-time keys" message, log in as an admin user and go to the landing page. You should see "Give one time keys" and if you click that box, the next message should be "Now you can give one time keys".

2. To test the "You are logged in to give one time keys", login as a staff user (or create one if you don't have one). If you log in as a staff user, the message you should see is "You are logged in to give one time keys"

### DIFF
--- a/covid_key/templates/covid_key/otk_start.html
+++ b/covid_key/templates/covid_key/otk_start.html
@@ -1,7 +1,13 @@
 {% extends 'base.html' %}
 {% load i18n %}
 
-{% block title %}{% trans "You’re logged in to give one-time keys" %}{% endblock %}
+{% block title %}
+  {% if request.user.can_send_alerts %}
+    {% trans "Now you can give one-time keys" %}
+  {% else %}
+    {% trans "You’re logged in to give one-time keys" %}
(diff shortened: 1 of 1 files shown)


covid_key/templates/covid_key/otk_start.html
```diff
@@ -4,7 +4,11 @@
 {% block title %}{% trans "You’re logged in to give one-time keys" %}{% endblock %}
```
dsamojlenko, line 4: Should probably switch the page title to match

sylviamclaughlin: Good idea! I will do it now and re-commit.

dsamojlenko (approved):

PR #665, opened 2021-06-29
Increase number of uwsgi worker processes

dsamojlenko:
# Summary | Résumé

In an effort to improve performance/stability of the container under load, increase the number of uwsgi workers (processes)


sylviamclaughlin (approved):
Looks good.

PR #673, opened 2021-07-02
Modify uwsgi threads/workers config

dsamojlenko:
# Summary | Résumé

- Breaks uwsgi config into .ini file for better readability
- Modifies threads/worker config to hopefully handle load better
- Increases unhealthy threshold and interval for Load Balancer health checks

The thinking here is that we know error rates were reduced significantly when adding workers. Adding workers also increased memory usage, so we ended up over the autoscale threshold and the service was permanently scaled up. Adding the worker scaling (cheaper config) should alleviate this as the task will start with a small number of workers and scale up under load.

We also know the Load Balancer is killing the ecs task because it sees it as unhealthy because it can't reach the health check. Setting the health check interval a bit longer and increasing the number of failed checks that would trigger an unhealthy status should give the task a bit of breathing room.

### DIFF
--- a/config/terraform/aws/lb_qrcode.tf
+++ b/config/terraform/aws/lb_qrcode.tf
@@ -12,13 +12,13 @@ resource "aws_lb_target_group" "qrcode" {
 
   health_check {
     enabled             = true
-    interval            = 10
+    interval            = 15
     path                = "/status/"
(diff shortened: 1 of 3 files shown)


sylviamclaughlin (approved):
Looks good to me.

Issue #681, opened 2021-07-06
Bug: Route53 health check failing for QR code English/French domains


patheard:
# Summary
With the addition of the Canada-only geolocation rule on the WAF, the DNS health checks are now failing for the QR code domains.  This is because the health checkers are all from regions outside of Canada.

A secondary bug is that the Staging Route53 health check for QR code is not attached to an alarm, so the failure was not noticed during testing.

PR #684, opened 2021-07-06
fix: remove QR code DNS health check

patheard:
# Summary
This will be replaced with an AWS CloudWatch Synthetic monitor because the Canada-only geolocation WAF rule blocks the Route53 DNS health checkers.

# Expected change
Remove QR code health check.

Related #681


dsamojlenko (approved):

patheard:
I'll fix up the VPN SG association and WAF PostRequestLimit rule once this finishes applying.

PR #686, opened 2021-07-07
feat: CloudWatch synthetic canary to monitor QR code

patheard:
# Summary
Creates a CloudWatch synthetic canary to check that the QR code `/status` endpoint is responding.  This change is required because the DNS health check we were using previously is being blocked by the WAF Canada-only geolocation rule.

Once this is working, a CloudWatch alarm will be added to check the canary response.

# Expected changes
1. New canary, S3 artifact bucket, execution role and IAM policy.
2. Revert WAF PostRequestLimit.
3. Revert VPN_SG attachment to database SG.

Related #681

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ secrets*.tfvars
 .terraform/
 *.tfplan
 .coverage
+node_modules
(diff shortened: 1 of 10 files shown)


dsamojlenko (approved):
cool

Issue #695, opened 2021-07-13
403 screen for geoblocking


YedidaZalik:
Please create the English and French screens as shown in figma:

English. https://www.figma.com/file/0CvVmmVb0EhnCmChzel70P/QR-Code-Registration-Website?node-id=4511%3A25651

French https://www.figma.com/file/0CvVmmVb0EhnCmChzel70P/QR-Code-Registration-Website?node-id=4554%3A25816

Merci!

sylviamclaughlin:
@YedidaZalik - what is the email address that I should put for the support email? Right now it is listed as support@email.ca. Thank you so much!

YedidaZalik:
That's a good question - I'm not sure.
@brianhendrick1 are you able to tell us?

PR #698, opened 2021-07-14
QR code 403 page changes

sylviamclaughlin:
# Summary | Résumé

Changes made as outlined in trello card https://trello.com/c/Hly46wM7/881-custom-403-page-for-geo-blocked-requests and  github ticket #695 

## Test instructions | Instructions pour tester la modification
Make sure that you change your environment variables to switch to the QR code application. Once you do that and restart the server, go to http://127.0.0.1:8000/403 to display the 403 page. The changes should be as outlined in the above referenced trello ticket and github issue. 
 
>


dsamojlenko:
Closing as this would not be the correct message if Django rendered the 403. This would need to be uploaded as a static HTML page to the WAF response for geo-blocked requests. Since we are not moving forward with this feature/site, let's just shelve it.

PR #734, opened 2021-08-09
bulk action created for admin

dorianjp:
* Added new bulk action for marking inactive
* Removed lock for sending surveys to health care users regardless whether they sent an alert previously or not.


sylviamclaughlin (approved):
Works great!

